PR #3, opened 2021-04-27
Documentation restructure

cristian-recoseanu:
Split documentation into 3 sections.
Added information on interoperability.

I have added the following constraints to IS-04 and IS-05 versions in the documentation:
* IS-04 APIs should be operating at version 1.2 or greater
* IS-05 APIs should be operating at version 1.1 or greater

Please review


N-Nagorny:
Looks good to me!

PR #7, opened 2021-05-12
Link definition of media profile in senders's behavior

prince-chrismc:
This is required to understand the meaning and implication of what the endpoint is enabled and how it should be used.
Highlighting it should make it easier for those reading and trying to follow.

### DIFF
--- a/docs/1.0. Overview.md
+++ b/docs/1.0. Overview.md
@@ -12,6 +12,6 @@ The terms ‘Device’, ‘Source’, ‘Flow’, ‘Sender’, ‘Receiver’ a
 
 A media consuming unit associated with an IS-04 Receiver. Sink is a resource and based on Resource Core JSON Schema.
 
-### Media Profile
+### Media Profiles
 
-Media description which describes a format acceptable for all Receivers implied to be a part of a connection. A Media Profile consist of separate parameters required to set up Sender and such IS-04 resources behind it as Source and Flow.
+A Collection of formats that are acceptable for all Receivers implied to be a part of a connection. A Media Profile consist of separate parameters required to set a Sender, Flow, and/or Source.
--- a/docs/1.0. Overview.md
+++ b/docs/1.0. Overview.md
@@ -14,4 +14,4 @@ A media consuming unit associated with an IS-04 Receiver. Sink is a resource and
 
 ### Media Profile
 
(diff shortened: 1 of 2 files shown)


docs/1.0. Overview.md
```diff
@@ -14,4 +14,4 @@ A media consuming unit associated with an IS-04 Receiver. Sink is a resource and
 
 ### Media Profile
 
-Media description which describes a format acceptable for all Receivers implied to be a part of a connection. A Media Profile consist of separate parameters required to set up Sender and such IS-04 resources behind it as Source and Flow.
+A Collection of formats that are acceptable for all Receivers implied to be a part of a connection. A Media Profile consist of separate parameters required to set a Sender, Flow, and/or Source.
```
N-Nagorny:
```suggestion
Description of a format that is acceptable for all Receivers implied to be a part of a connection. A Media Profile consist of separate media parameters required to set a Sender, Flow, and/or Source.
```
It is a single format, not a collection.

prince-chrismc: :confused: https://github.com/AMWA-TV/nmos-sink-metadata-processing/blob/75e292aa65a28d63a057a0fded694cb0d30a47f8/APIs/schemas/media_profiles.json#L3

docs/1.2. Behaviour.md
```diff
@@ -4,7 +4,7 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
 
 ## Senders
 
-`/media-profiles` allows the client to pass an array of Media Profiles to a Sender and let a third-party which controls it choose what exact format to use. A client can also pass a single Media Profile to specify the exact format.
+`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the device which controls it choose what exact format to use. A client can also pass a single Media Profile to specify the exact format.
```
N-Nagorny: `Device` is quite physical. I'm not sure if we should avoid terms which are 100 % hardware or software related. If we should, then I'd propose `underlying component`.

prince-chrismc: Hows this?

```suggestion
`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component control which it choose exactly which format to use. A client can also pass a single Media Profile to specify the exact format.
```

PR #12, opened 2021-05-21
A Sender MUST obey it's Media Profile

### LINKED ISSUE #15: /media-profiles PUT operation description is vague
I understood my concerns regarding https://github.com/AMWA-TV/nmos-sink-metadata-processing/commit/711b770fb1e5d2b4a1b7ef89598da07fed038e7d. The PUT operation carries a set of Media Profiles which doesn't fit `if the Device supports the proposed media profile`.

Implementations may take this set, validate it against the schema (return 400 if error), then let a component behind the Sender to choose supported ones and return them with 200 or return 422 if no one is supported.

prince-chrismc:
resolves #15


docs/1.2. Behaviour.md
```diff
@@ -8,15 +8,19 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
 
-A `DELETE` request MUST clear the last successfully applied Media Profiles array.
+If the new `/media-profiles` structure causes underlying IS-04 flows and sources to change, then their version MUST be updated (in registered mode this MUST update the registered resources).
 
-When the `media-profiles` structure changes as a consequence of PUT or DELETE (request is validated and accepted by the API server), then the associated IS-04 Sender MUST have its version updated (in registered mode this MUST update the registered resource).
-
-If the new `media-profiles` structure causes underlying IS-04 flows and sources to change, then their version MUST be updated (in registered mode this MUST update the registered resources).
+When an external change (i.e. not from the `/media-profiles` endpoint) modifies the associated Source and/or Flow, the Sender MUST respect its
+[Media Profiles](1.0.%20Overview.md#media-profile). If the change does not satisfy its Media Profiles the Sender MUST NOT transmit the signal over the network.
```
N-Nagorny:
```suggestion
When an external change (i.e. not from the `/media-profiles` endpoint) modifies a Source and/or a Flow of a Sender and/or the Sender itself, it MUST respect [Media Profiles](1.0.%20Overview.md#media-profile) of this Sender. If the change does not satisfy its Media Profiles the Sender MUST NOT transmit the signal over the network.
```

docs/1.2. Behaviour.md
```diff
@@ -8,15 +8,19 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
+The Sender is REQUIRED to take the following actions:
+- MUST set the IS-05 `/active`'s `master_enabled` property to `false`.
```
N-Nagorny:
```suggestion
- MUST set the IS-05 `/active`'s `master_enable` property to `false`.
```

N-Nagorny (commented):
Looks good.

docs/1.2. Behaviour.md
prince-chrismc: Maybe just "any change" not simply those covered currently?

```suggestion
When an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream then the Sender MUST respect its
[Media Profiles](1.0.%20Overview.md#media-profile). If the change does not satisfy its Media Profiles the Sender MUST NOT transmit the signal over the network.
```

N-Nagorny: Sounds appropriate. The only thing which seems possibly wrong is that "Sender MUST respect Media Profiles". We can say the change must respect them or use passive voice.

N-Nagorny:
Is that self-explanatory enough if a component behind Sender goes out of the Media Profiles, it's reflected in the Flow and  `subscription` of the Sender becomes clear? Can we signalize somehow why transmitting is stopped?

N-Nagorny:
If subscription emptied when Media Profiles rule is broken, should IS-05 deny further connections until the Flow matches the rule again? If yes, how to handle scheduled activation?

prince-chrismc:
Excellent questions.

> if a component behind Sender goes out of the Media Profiles, it's reflected in the Flow and `subscription` of the Sender becomes clear?

The media profile maps to the IS-04 resources and it's described right above https://github.com/AMWA-TV/nmos-sink-metadata-processing/blame/60f8efebf1f31a87ea5f85bd8281b104593503e3/docs/1.2.%20Behaviour.md#L17
I think it's a little much, but I have no strong feeling on removing it.

> Can we signalize somehow why transmitting is stopped?

In our dreams 🙃 "why" is a hard thing to push with what we currently have

> If subscription emptied when Media Profiles rule is broken, should IS-05 deny further connections until the Flow matches the rule again? If yes, how to handle scheduled activation?

As I've understood things, we have

1. Flow changing outside of media profiles
    a. Subscription active set to false
    b. follow IS-04 rules
2. Controller picks up that changes and does IS-05 to correct the situation (optional)

I think this is an excellent point. 

❓ Can we active a sender via IS-05 who's stream does **not** respect the media profiles?

N-Nagorny:
> Can we active a sender via IS-05 who's stream does **not** respect the media profiles?

My answer is _no_. If a user wants to do this, they should DELETE the last Media Profiles which turns off IS-11.

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,31 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
 
 ## Senders
 
-`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component control which it choose exactly which format to use. A client can also pass a single Media Profile to specify the exact format.
+The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component to choose exactly which format from the set it will use. A client can pass a single Media Profile to specify the exact format.
```
N-Nagorny:
```suggestion
The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying implementation to choose exactly which format from the set it will use. A client can pass a single Media Profile to specify the exact format.
```

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,31 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
 
-The initial state of Media Profiles of any Sender is empty. Creating a connection with such a Sender via IS-05 doesn't involve the Sink Metadata Processing API.
+The initial state of Media Profiles of a Sender MUST be an empty JSON array `[]`. This state indicated a Sender is unbound by an Media Profile. Creating a connection with such a Sender via [IS-05][IS-05] is not impacted by the Sink Metadata Processing API.
```
N-Nagorny:
```suggestion
The initial state of Media Profiles of a Sender MUST be an empty JSON array `[]`. This state indicated a Sender is unbound by a Media Profile. Creating a connection with such a Sender via [IS-05][IS-05] is not impacted by the Sink Metadata Processing API.
```

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,31 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
+- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
+- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
 
-A `PUT` request MUST be validated and if the Device supports the proposed media profile, it initiates an attempt to reconfigure the Device and update corresponding Flow, Source, and/or Sender properties. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.
+If a Sender, its associated Flow, and/or Source do not meet the restrictions of a the Media Profiles, the Sender MUST refuse an [IS-05][IS-05] `/active` requests which tries to set `master_enable` property to `true` by responding with a `409 Conflict` response.
```
N-Nagorny:
```suggestion
If a Sender, its associated Flow, and/or Source do not meet the restrictions of the Media Profiles, the Sender MUST refuse an [IS-05][IS-05] `/staged` requests which tries to set `master_enable` property to `true` by responding with a `409 Conflict` response.
```

N-Nagorny:
Find an appropriate error message for the conflict case caused by media format unsupported by Media Profiles.

N-Nagorny:
Think about Media Profile usage status (which shows whether its policy is broken).

N-Nagorny:
Is it an appropriate solution to check whether a Flow matches Media Profiles on the Controller side? Controller tracks Senders via polling Nodes or Query API subcription. If a Sender unexpectedly stops, it may get Flow and Media Profiles and do the check.

prince-chrismc:
There was two big concerns after the last call

1. Can blocking IS-05 requests be adequately tied to IS-11?
2. How does a controller know the Media Profile is in effect/active?

One of the ideas, and the proposed solution, is to _clear_ the Media Profiles when an external change forces the Sender to not conform to the set.

When the Media Profile is `[]` it has an exact definition which is closely related to the lack of conformity from the use-case in question.

N-Nagorny:
Cleaning Media Profiles is an action which can be performed via the API. Such properties doesn't seem a good indicator because it's not clear whether a Controller stopped a stream and cleared the Media Profiles or the API implementation did this to show that Flow is outside of the Media Profiles.

If `/media-profiles` accepts any Media Profiles valid against the schema and guarantees that the Flow matches them, then there is no way for Sender to say it doesn't support some Media Profiles. If the Controller can't ask the Sender what it supports, the it only can check the actual Flow. If the Controller notices transmitting interruption, it can check the Flow and it needs the Media Profiles to check whether the Flow matches them.

We don't know what the user can effect after the interruption. They may manage Sources out of NMOS. In this case they can look at the preserved Media Profiles and change the Source to one which has an appropriate Flow. They also may want to find a Receiver which doesn't support the actual Flow, turn it off, expand Media Profiles and re-enable transmitting (this case isn't a strong reason because the Controller can build Media Profiles from the scratch).

prince-chrismc:
🤔 All three of your points I would put under "Persistence of Media Profiles" chapter. Which we haven't discussed in the "Architecture" or design....

***

For the 1st paragraph, this is what we have in production for IS-05 workflows. A controller (most often) keeps an internal database with the state it needs or the user selected.
I would not expect a controller does not track changes to the media profile. It will only GET the _current state_ regardless of who or what set that. Making changes it requires without any prior knowledge of the history.

As you noted before, we could remove the Media Profiles which are no supported but that will limit the devices ability to handle external changes.
For instance a Sendering following exactly the source, will known **only one** profile is acceptable but it will not be able to verify the rest.

I :100: agree stopping the stream is sufficient to indicate an issue, however we hit the issue https://github.com/AMWA-TV/nmos-sink-metadata-processing/pull/12#issuecomment-852438284 about mismatching IS-11 and IS-05.

It's not (reasonable) possible to indicate to the controller when the IS-05 is blocked by IS-11... which was the last idea.
We want/need existing IS-05 based controllers to be able to operator with us, so we can not extend the error handling just for IS-11.

***

What I expect to be the workflow for 🎮 Controllers

is-11 (1)

This matches my experience with IS-05 controllers

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -24,6 +24,8 @@ mediaType: application/json
           body:
             example: !include ../examples/sender-get-200.json
             type: !include schemas/edidapi-sender.json
+        204:
+          description: Denotes an external change occured which violated the previous Media Profiles.
```
prince-chrismc:
```suggestion
        205:
          description: Denotes an external change occurred which violated the previous Media Profiles.
```

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,37 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
+Once a Media Profile set is accepted, a Sender MUST transmit a Flow and Source which meets at least one profile. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy its Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
 
-A `DELETE` request MUST clear the last successfully applied Media Profiles array. It turns Media Profiles to the initial state which exclude Sink Metadata Processing out of the process.
+- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
+- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
+- MUST respond with `204 No Content` when answering `GET /media-profiles` to indicate it is no longer following or adhering to a set. This response SHOULD presist until the next valid `PUT /media-profiles` request.
```
prince-chrismc:
```suggestion
- MUST respond with `205 Reset Content` when answering `GET /media-profiles` to indicate it is no longer following or adhering to a set. This response SHOULD persist until the next valid `PUT /media-profiles` request.
```

prince-chrismc (commented):
I am coming back with another proposal....

Using the `GET` response to  indicate something went wrong

N-Nagorny:
It looks good. Let me describe some scenarios as I understand them to make sure we're on the same page.

Flow hereinafter means a Flow associated with a Sender that accompanied with Media Profiles.

Scenario 1. A client initiates transmitting
The Node checks if a Flow matches the Media Profiles. If it does not, then IS-05 returns an error.

Scenario 2. A Flow is changed during transmitting
If a Flow is changed, then the Node checks if the new Flow matches the Media Profiles. If it does not, then the Node stops transmitting.

Scenario 3. A client puts Media Profiles while transmitting
We can't guarantee that IS-11 implementation will update Flow immediately after putting Media Profiles. Should we prohibit this scenario?

prince-chrismc:
> Scenario 1. A client initiates transmitting
> The Node checks if a Flow matches the Media Profiles. If it does not, then IS-05 returns an error.

A Flow SHOULD never be outside of the of the Media Profiles. When a Flow does not satisfy its Media Profiles, the implementation MUST clearly indicate that... I think there's a tweak to the current PR for that.

> Scenario 2. A Flow is changed during transmitting
> If a Flow is changed, then the Node checks if the new Flow matches the Media Profiles. If it does not, then the Node stops transmitting.

This is covered to my satisfaction

> Scenario 3. A client puts Media Profiles while transmitting
> We can't guarantee that IS-11 implementation will update Flow immediately after putting Media Profiles. Should we prohibit this scenario?

There's no "response not ready" that's supported in the the `PUT` or `DELETE` so I was under the impression it MUST be completed before responding

N-Nagorny:
I don't see obstacles to return `423 Locked` from `PUT` and `DELETE` if an associated Sender is active. I think, we need to cover the case when the Node can't provide a user with a matching Flow.

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,37 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
+### Role of Media Profiles
 
-A `GET` request returns the last successfully applied Media Profiles array.
+The initial state of Media Profiles of a Sender MUST be an empty JSON array `[]`. This state indicated a Sender is unbound by a Media Profile. Creating a connection with such a Sender via [IS-05][IS-05] is not impacted by the Sink Metadata Processing API.
 
-A `PUT` request MUST be validated and if the Device supports the proposed media profile, it initiates an attempt to reconfigure the Device and update corresponding Flow, Source, and/or Sender properties. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.
+Once a Media Profile set is accepted, a Sender MUST transmit a Flow and Source which meets at least one profile. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy its Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
```
prince-chrismc:
```suggestion
Once a Media Profile set is accepted, a Sender MUST have a Flow and Source which meets at least one profile. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy its Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
```

N-Nagorny:
After the call we had today, my understanding of the process is the following:
1. At start Media Profiles = `[]` which means IS-11 is off.
2. Each PUT tries to modify the Source (if audio), the Flow and the Sender (updates its version) to satisfy the Media Profiles. If it is not possible, then an error returned and nothing changed (if IS-11 was off it stays off, if there were different Media Profiles they stay active).
3. If external Flow change violates the Media Profiles, the Sender stops transmitting and GET returns a specific error code which means IS-11 terminated unsuccessfully.

If I didn't miss anything, then the actual state of the branch covers these items.

Regarding the error code, 205 now looks appropriate to me. "No Content" is very close to `[]`, and the meaning doesn't directly follow from the response. 205 is farther from `[]` than 204 and it's closer to the recommendation of performing DELETE Media Profiles or PUT new ones.

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,39 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
 
-A `DELETE` request MUST clear the last successfully applied Media Profiles array. It turns Media Profiles to the initial state which exclude Sink Metadata Processing out of the process.
+- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
+- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
+- MUST respond with `204 No Content` when answering `GET /media-profiles` to indicate it is no longer following or adhering to a set. This response SHOULD presist until the next valid `PUT /media-profiles` request.
```
prince-chrismc:
```suggestion
- MUST respond with `205 Reset Content` when answering `GET /media-profiles` to indicate it is no longer following or adhering to a set. This response SHOULD persist until the next valid `PUT /media-profiles` request.
```

docs/1.2. Behaviour.md
```diff
@@ -4,30 +4,39 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
...
 
-When the `media-profiles` structure changes as a consequence of PUT or DELETE (request is validated and accepted by the API server), then the associated IS-04 Sender MUST have its version updated (in registered mode this MUST update the registered resource).
+When a Sender is in this violated state, responding with a `204 No Content`, This state indicated a Sender is unbound by a Media Profile and is similar to the initial state.
```
prince-chrismc:
```suggestion
When a Sender is in this state, ie. responding with a `205 Reset Content`, this indicates a Sender is unbound by a Media Profile and is similar to the initial state.
```

prince-chrismc:
We are still considering the following mechanisms:

1. Returned a success code to indicate the violated Media Profiles
2. Exposing a new attribute (same or new endpoint) to show where Media Profiles are "in use"

prince-chrismc:
> 2\. Exposing a new attribute (same or new endpoint) to show where Media Profiles are "in use"

The only technical reason against this is the media profiles is invalidate and it should not be exposed in that state (it no longer carries any meaning). For this reason I have a slight preference for using the status code

PR #13, opened 2021-06-01
Describe format change beyond NMOS

N-Nagorny:
It is possible that the underlying component behind a Device with a Sender changes format on its own. This PR adds a requirement for changes to stay within Media Profiles set.

PR #14, opened 2021-06-01
Comment DELETE operation

N-Nagorny:
Added a comment saying that DELETE turns off Sink Metadata Processing.

### DIFF
--- a/docs/1.2. Behaviour.md
+++ b/docs/1.2. Behaviour.md
@@ -4,7 +4,7 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
 
 ## Senders
 
-`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component control which it choose exactly which format to use. A client can also pass a single Media Profile to specify the exact format.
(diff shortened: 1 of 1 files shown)


docs/1.2. Behaviour.md
```diff
@@ -4,15 +4,15 @@ The Sink Metadata Processing API provides a mechanism to change the settings ass
 
 ## Senders
 
-`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component control which it choose exactly which format to use. A client can also pass a single Media Profile to specify the exact format.
+`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender, let the underlying component choose exactly which format to use at the moment and dictate that format changes which are not consequent to Media Profiles change MUST lay within Media Profiles set. A client can also pass a single Media Profile to specify the exact format.
```
prince-chrismc: This is from #13 ?

N-Nagorny: Yes. I thought it's just extra to #13 but it may be redundant.

N-Nagorny:
```suggestion
`/senders/{senderId}/media-profiles` allows the client to pass an array of [Media Profiles](1.0.%20Overview.md#media-profile) to a Sender and let the underlying component control which it choose exactly which format to use. A client can also pass a single Media Profile to specify the exact format.
```

N-Nagorny: Moved this out.

PR #16, opened 2021-06-01
Split media_profile.json into separate video and audio flavors

N-Nagorny:


peterbrightwell:
LGTM

cristian-recoseanu (commented):
It looks good. Thanks @N-Nagorny !

cristian-recoseanu (approved):

prince-chrismc (approved):
🚀

APIs/schemas/media_profile_audio.json
```diff
@@ -0,0 +1,69 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "grain_rate" : {
```
prince-chrismc, line 5: This

APIs/schemas/media_profile_audio.json
```diff
@@ -0,0 +1,69 @@
...
+          "type" : "integer"
+        },
+        "denominator" : {
+          "description" : "Denominator",
+          "type" : "integer",
+          "default" : 1
+        }
+      }
+    },
+    "sample_rate" : {
```
prince-chrismc, line 23: vs this

prince-chrismc (commented):
https://specs.amwa.tv/bcp-004-01/releases/v1.0.0/examples/receiver-audio.html

Is "sample" ... not sure that the difference here

cristian-recoseanu:
> 
> 
> https://specs.amwa.tv/bcp-004-01/releases/v1.0.0/examples/receiver-audio.html
> 
> Is "sample" ... not sure that the difference here

That's a very good question.

Looking at the BCP-004-01 parameter register it seems both are valid parameter constraints

https://specs.amwa.tv/nmos-parameter-registers/branches/main/capabilities/

But then looking at the examples it seems that grain_rate is used for video streams and sample_rate for audio streams.

prince-chrismc:
```
Name: urn:x-nmos:cap:format:grain_rate
Description: Identifies streams which may be created or consumed based upon their grain rate.
Specification: per AMWA BCP-004-01
Type: rational
Target: (a) Flow ‘grain_rate’ (or from Source if omitted), (b) SDP attribute ‘a=fmtp:’ format-specific parameter ‘exactframerate’
```

```
Name: urn:x-nmos:cap:format:sample_rate
Description: Identifies the acceptable number of samples per second in an audio stream.
Specification: per AMWA BCP-004-01
Type: rational
Target: (a) audio Flow ‘sample_rate’, (b) SDP attribute ‘a=rtpmap:’ <clock rate>
```

I think it's sample 🤔

N-Nagorny:
`A group of consecutive audio samples` is given as an example of an audio Grain.
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/5.0.%20Data%20Model.md#sources-flows-and-grains

Does it make sense to remove `grain_rate` from `media_profile_audio.json`? I never seen any example of controlling audio Flows on the audio Grain level.

prince-chrismc:
:wave: @garethsb @andrewbonney Should "raw audio flow" use `grain_rate` or `sample_rate` or both? It's a bit unclear for us (thanks in advance)

garethsb:
`sample_rate` is well defined for audio Flows, `grain_rate` for audio Sources and Flows less so... See e.g. https://github.com/AMWA-TV/nmos-testing/issues/517 for some of the previous discussion.

garethsb:
Several possibilities for `grain_rate`:
- Set it to the same as the `sample_rate` in the Flows (and to meet the requirements, therefore to the LCM of the Flows' value in the Source).
- Set it in the Flow based on the GCD of the packet time of the streams of the Senders (and as above in the Source).
- Set it to match the related video Flows and Source if they derive from a common mux, e.g. SDI.

I'm not aware of a good reason to favour one approach rather than another (is the additional information provided by any of these approaches useful?) and different vendors have done different things; the testing tool no longer checks the specific value. Specifying receiver capabilities for audio Receivers based on `grain_rate` therefore also seems unwise.

prince-chrismc:
Since this is new let's try to constrain things? Its fairly messy 🤧 

We can allow both in the schema and add a test for unique entries or the same value?
Could we specify it with normative language?

N-Nagorny:
I'll extract the `grain_rate` question into a separate issue. Now merging in what we have now.

PR #25, opened 2021-06-14
Add missing response body

prince-chrismc:


prince-chrismc:
❓ Should a `DELETE` return the empty Media Profiles?

N-Nagorny:
It doesn't make sense for me because it doesn't give any additional information. Will it be more consistent if do so?

prince-chrismc:
> It doesn't make sense for me because it doesn't give any additional information.

I agree, and it there is no comparison for consistency

PR #26, opened 2021-06-15
Remove grain_rate from media_profile_audio.json

### LINKED ISSUE #22: Should `grain_rate` be removed from `media_profile_audio.json`?
There is `grain_rate` property which is used in both video and audio Flows. Therefore, it's present in the Media Profile schema of both types.
The discussion is started in #16.

N-Nagorny:
Resolves #22.


prince-chrismc:
Could we sneak in an audio example?

prince-chrismc (approved):

prince-chrismc:
Please link #22 which should close with this PR :)

PR #28, opened 2021-06-17
Add "required" to Media Profiles Schema

### LINKED ISSUE #27: Making Media Profiles from Receiver Caps is not defined well
Receiver Caps is more flexible than Media Profiles, and when we prescribe making the second from the first, it raises question "If Receiver Caps don't constrain a parameter or do express a constraint with a range, should Controller create Media Profiles for all possible combinations of parameters matching the constraints?"

```
|                  Sender                   |                      |                Receiver                |
| HDMI Source <--> EDID <--> Media Profiles | <--> Controller <--> | Receiver Caps <--> EDID <--> HDMI Sink |
```

My proposal is to:
1. Expand IS-11 with a statement prescribing that Receivers listed in IS-11 MUST follow BCP-005-01.
2. Explicitly require `enum` usage for describing timings in BCP-005-01.
3. Add `required` fields to Media Profiles and state that absent fields mean any value of a parameter.

N-Nagorny:
Resolves #27.

### DIFF
--- a/APIs/schemas/media_profile_audio.json
+++ b/APIs/schemas/media_profile_audio.json
@@ -2,7 +2,8 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "required": [
-    "media_type"
+    "media_type",
+    "channel_count"
   ],
   "properties": {
     "sample_rate" : {
--- a/APIs/schemas/media_profile_audio.json
+++ b/APIs/schemas/media_profile_audio.json
@@ -1,6 +1,9 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
+  "required": [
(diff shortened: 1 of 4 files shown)


prince-chrismc (approved):

cristian-recoseanu:
I think we should discuss this a bit more. Anything you make required in the schema has implications on BCP-005-01 because it means receivers are mandated to have some sort of information regarding that parameter in the constraints.
This means receivers cannot leave the mandatory parameters unconstrained as this schema does not allow missing parameters (meaning unconstrained).

prince-chrismc (approved):

prince-chrismc (commented):
This does not seem to cover the linked issue, is there some text that is missing?

prince-chrismc (approved):

PR #35, opened 2021-06-24
Add preference to Media Profiles

### LINKED ISSUE #31: Media Profiles and Prefered Mode
One important attribute in an EDID is the prefered mode. We have documented this in bcp-005-01 but it does not carry over to the Media Profiles.

When a `PUT /media-profiles` occurs how does the sender know which corresponds to the "prefered mode"?

N-Nagorny:
Resolves #31.

### DIFF
--- a/APIs/schemas/media_profile_audio.json
+++ b/APIs/schemas/media_profile_audio.json
@@ -2,6 +2,11 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
+    "preference": {
+      "type": "number",
+      "maximum": 100,
+      "minimum": -100
+    },
     "sample_rate" : {
       "description": "Number of audio samples per second for this Flow",
       "type": "object",
--- a/APIs/schemas/media_profile_audio.json
+++ b/APIs/schemas/media_profile_audio.json
@@ -4,6 +4,7 @@
   "properties": {
     "preference": {
       "type": "number",
+      "default": 0,
       "maximum": 100,
       "minimum": -100
     },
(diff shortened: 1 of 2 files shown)


cristian-recoseanu:
Don't we need to have preference for both the Video and Audio variant of the media profile?

N-Nagorny:
I didn't find preferred audio modes in BCP-005-01 and the EDID specification.

prince-chrismc (approved):

cristian-recoseanu:
So just to clarify does the preference property in the Media Profile not get filled in from receiver capabilities? BCP-004-01 supports that for anything. So we could have audio receivers which have preferences in receiver capabilities. Is this saying we should ignore preference from audio receivers?

APIs/schemas/media_profile_audio.json
```diff
@@ -2,6 +2,11 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
+    "preference": {
+      "type": "number",
```
prince-chrismc, line 6:
```suggestion
      "type": "number",
      "default": 0,
```

prince-chrismc (commented):
From our call

APIs/schemas/media_profile_video.json
```diff
@@ -2,6 +2,11 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
+    "preference": {
+      "type": "number",
```
N-Nagorny, line 6:
```suggestion
      "type": "number",
      "default": 0,
```

PR #41, opened 2021-06-28
Make /sinks/{sinkId}/edid discoverable and pull up properties under "…

### LINKED ISSUE #17: Should /sinks have similar subroutes to /senders and /receivers?
`/senders/`/ `/receivers/` + `resource ID` return a subroute. `/sinks/` + `Sink ID` returns a Sink JSON, hiding `/edid` subroute. Should we remake this in the consistent way?

N-Nagorny:
…edid" in Sink schema

Resolves #17 and removes `edid` property from Sink schema since EDID may be absent.


examples/sender-get-200.json
```diff
@@ -1,3 +1,3 @@
 [
-  "media-profiles/"
+  "properties/"
```
prince-chrismc: This looks out of place 🤔

prince-chrismc:
Unrelated but I just noticed in the other APIs when there's an "id" in the path the subsequent endpoints have 404 responses documented https://github.com/AMWA-TV/nmos-device-connection-management/blob/ada62dbcc023a0132eddfd9c59cd00ab8a236372/APIs/ConnectionAPI.raml#L199-L200

We are missing those in our RAML

examples/sender-get-200.json
prince-chrismc:
```suggestion
  "media-profiles/"
```

prince-chrismc (commented):
I think you modified one file by accident

N-Nagorny:
@prince-chrismc, thank you for the catch! I committed your fix and added 404.

prince-chrismc (approved):

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -117,22 +125,40 @@ mediaType: application/json
...
-      description: Return Sink.
+      description: List all the API endpoints for a single Sink.
       responses:
         200:
           description: Returned when success.
           body:
             example: !include ../examples/sink-get-200.json
-            type: !include schemas/sink.json
-    /edid:
+            type: !include schemas/sink-api-sink.json
```
garethsb, line 134: FYI, @N-Nagorny, several other specs call these schemas like `base`, e.g. `schemas/sink-base.json` in this case.

PR #61, opened 2021-07-30
Describe the case when Sender can't apply Media Profiles immediately

### LINKED ISSUE #58: Time between passing EDID to HDMI Source and InfoFrame change is unknown
Now PUT `/senders/{senderId}/media-profiles` is an atomic operation which immediately returns a response showing whether Media Profiles were applied successfully. In terms of HDMI it's needed to build an EDID from Media Profiles and pass it to HDMI Source. Then it changes video format and notifies about it with a new InfoFrame. The problem is that we don’t know time between passing EDID to HDMI Source and new InfoFrame appearance which may lead us to another PUTting approach.

A possible solution is prohibiting PUTting while streaming for implementations which can't handle such change on the fly. In this case PUT response code is 202 instead of 200 which means the change wasn't applied immediately and Controller needs to wait for a Flow change. After the change Controller GETs `/senders/{senderId}/media-profiles` (whether it's 205).

N-Nagorny:
Resolves #58.

### DIFF
--- a/docs/2.1. Behaviour - Server Side.md
+++ b/docs/2.1. Behaviour - Server Side.md
@@ -12,15 +12,13 @@ The initial state of Media Profiles of a Sender MUST be an empty JSON array `[]`
 
 Non-empty array of Media Profiles MUST be treated as unordered unless `preference` property exists.
 
-Once an array of Media Profiles is accepted, a Sender MUST have a Flow and Source which meet at least one profile. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy the Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
+Once `PUT /media-profiles` is requested, the Node MUST respond and the Sender MUST change the Flow and/or the Source to meet at least one profile as soon as possible. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy the Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
 
 - MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
 - MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
 - MUST respond with `205 Reset Content` when answering `GET /media-profiles` to indicate it is no longer following or adhering to a set. This response SHOULD persist until the next valid `PUT /media-profiles` request.
 
 When a Sender is in this state, i.e. responding with a `205 Reset Content`, this indicates a Sender is unbound by Media Profiles and is similar to the initial state.
-
-If Sender is unable to apply requested Media Profiles immediately, the Node should respond with `202 Accepted` which signalizes that Sender will apply the changes as soon as possible, update Source and Flow accordingly and indicate about unbounding by Media Profiles if the resulting change breaks them.
 
 ### Request Methods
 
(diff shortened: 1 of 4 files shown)


docs/2.1. Behaviour - Server Side.md
```diff
@@ -12,13 +12,13 @@ The initial state of Media Profiles of a Sender MUST be an empty JSON array `[]`
 
 Non-empty array of Media Profiles MUST be treated as unordered unless `preference` property exists.
 
-Once an array of Media Profiles is accepted, a Sender MUST have a Flow and Source which meets at least one profile. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy its Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
+Once `PUT /media-profiles` is requested, the Node MUST respond and the Sender MUST change the Flow and/or the Source to meet at least one profile as soon as possible. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy the Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
```
prince-chrismc:
```suggestion
Once `PUT /media-profiles` is requested, the Node MUST validate the request before responding. The Sender MUST change the Flow and/or the Source to meet at least one profile as soon as possible. This constraint MUST be adhered to when an external change (i.e. not from the `/media-profiles` endpoint) modifies any aspect of the stream. If the change does not satisfy the Media Profiles, the Sender MUST NOT transmit the signal over the network. The Sender is REQUIRED to take the following actions:
```

N-Nagorny: Validating should be mentioned, but what is "source"? What if the underlying implementation is software only?

prince-chrismc: The IS-04 definition is a Source is the origin of the content... which I assume is what needs the EDID to change the resolution regardless of the implementation

N-Nagorny: I may be wrong, but in my understanding [a logical Source](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/5.0._Data_Model.html#sources-flows-and-grains) isn't required to represent an HDMI Source which needs the EDID. I don't think we can present something to "an abstract logical point", it's just a representation of something real which we should communicate to. After we communicated to it, the logical Source reflects changes.

prince-chrismc: I removed the source part to avoid confusion

prince-chrismc (approved):

prince-chrismc:
❓ should we allow the `200` for everything meets media profiles?
- no changes required
- all adjustments have been made

N-Nagorny:
Let's consider two examples.

A. Sender must change Flow after receiving Media Profiles.

1. HDMI Source gives 1080p video, Media Profiles are empty.
2. Media Profiles excluding 1080p passed, 202 returned.
3. Gateway Sender modifies its EDID.
4. HDMI Source handles the EDID change.
5. Flow changes if HDMI Source starts giving another video format.

B. Sender doesn't have to change Flow after receiving Media Profiles.

1. HDMI Source gives 1080p video, Media Profiles are empty.
2. Media Profiles including 1080p passed, 200 returned.
3. Gateway Sender modifies its EDID anyway.
4. HDMI Source handles the EDID change anyway which may cause Flow change.
5. Flow changes if HDMI Source starts giving another video format for some reason.

It makes sense in general to say "Media Profiles accepted, Sender satisfies them" but it doesn't equal to "No changes needed" and Node can't guarantee Flow won't be changed. Or you mean to return 200 if passed Media Profiles are equal to the current ones?

N-Nagorny:
Closed on the group meeting on January 24, 2022.

PR #62, opened 2021-08-13
Use docs/README.md instead of document numbers

peterbrightwell:
We are migrating from using `1.0.` etc to determine the document ordering and levels to specifying this directly in `docs/README.md`


prince-chrismc (approved):
Most of the links pointed outside the spec 👍

N-Nagorny (approved):

PR #65, opened 2021-10-18
Move from Sinks to Inputs and Outputs

N-Nagorny:


APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -104,45 +115,93 @@ mediaType: application/json
...
+      get:
+        description: Return properties of the Input.
+        responses:
+          200:
+            description: Returned when success.
+            body:
+              example: !include ../examples/input-properties-get-200.json
+              type: !include schemas/input.json
+          404:
+            description: Returned when the requested Sink ID does not exist.
```
alabou: "Sink ID" to be replaced by "Input ID"

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -104,45 +115,93 @@ mediaType: application/json
...
         responses:
           200:
             description: Returned when success.
             body:
-              example: !include ../examples/sink-properties-get-200.json
-              type: !include schemas/sink.json
+              example: !include ../examples/output-properties-get-200.json
+              type: !include schemas/output.json
           404:
             description: Returned when the requested Sink ID does not exist.
```
alabou: "Sink ID" to be replaced by "Output ID"

docs/Data Model.md
```diff
@@ -9,11 +9,18 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 ![Extended IS-04 Model Diagram](../docs/images/is-11-data-model.png)
 
-The IS-04 heirarchy is preserved with the exception of the addition of a Receiver can identify an association with 0 or more Sink resources.
+The IS-04 heirarchy is preserved with the exception of the addition of Sender can identify an association with 0 or more Input resources and Receiver can identify an association with 0 or more Output resources.
```
alabou, line 12: hierarchy

docs/Behaviour - Server Side.md
```diff
@@ -1,18 +1,20 @@
 # API
 
-The Sink Metadata Processing API provides a mechanism to change the settings associated with logical Sources, Flows and Senders. Another use case is obtaining information about Sinks associated with Receivers.
+The Sink Metadata Processing API provides a mechanism to change the settings associated with logical Inputs, Sources, Flows and Senders. Another use case is obtaining information about Outputs associated with Receivers.
 
 ## Senders
 
-The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](Overview.md#media-profile) to a Sender and let the underlying implementation to choose exactly which format from the set it will use. A client can pass a single Media Profile to specify the exact format.
+The `/outputs` endpoint returns a list of IDs corresponding to Inputs associated to the specific Sender. If this list changes then the associated [IS-04][IS-04] Sender MUST update its version (in registered mode this MUST update the registered resource).
```
alabou: The `/outputs` endpoint => The `/inputs` endpoint

docs/Data Model.md
```diff
@@ -9,11 +9,18 @@ In order to ensure consistent behaviour in a facility employing this identity, s
...
```
alabou: hierarchy

docs/Data Model.md
```diff
@@ -9,11 +9,18 @@ In order to ensure consistent behaviour in a facility employing this identity, s
...
 
-## Sink Representation
+## Input Representation
 
-A new Sink is created in the following circumstance:
+A new Input is created in the following circumstance:
+
+- At the presence of an IS-11 compatible Sender
```
alabou: Not sure what this mean ... this first lie is different from the Receiver side ...

docs/Data Model.md
```diff
@@ -24,26 +31,47 @@ As per IS-04,
 
 > Persistent identity helps to ensure that users’ expectations meet with reality. For example, if a piece of equipment is switched off and back on, the user would expect it to return with ... the identifiers used by the Node remain the same.
 
-### Sink ID
+### Input ID
+
+Owned by:
+
+- Sender ID
```
alabou: I'm not sure I agree with this ... An input shall be associated to a device.

docs/Data Model.md
```diff
@@ -24,26 +31,47 @@ As per IS-04,
...
+
+- The parent Sender ID
+- A connection identifier such as a channel name, index or similar
+
+### Output ID
 
 Owned by:
 
-- Device ID
+- Receiver ID
```
alabou: I'm not sure I agree with this ... An output shall be associated to a device.

docs/Overview.md
```diff
@@ -9,7 +9,7 @@
 
 ## Introduction
 
-AMWA IS-11 specifies how how Sources, Flows and Senders within an NMOS compatible system can be reconfigured and EDIDs of Sinks associated with Receivers can be obtained.
+AMWA IS-11 specifies how Sources, Flows and Senders within an NMOS compatible system can be reconfigured and EDIDs of Outputs associated with Receivers can be obtained.
```
alabou: This should also state that receiver capabilities are required, that effective EDID of input can be obtained, etc ...

docs/Overview.md
```diff
@@ -31,13 +31,17 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
 
 The terms 'Node', 'Device', 'Sender' and 'Receiver' are used extensively in this documenation set. The [NMOS Technical Overview](https://specs.amwa.tv/nmos/main/docs/2.0._Technical_Overview.html) provides an outline of these terms, and IS-04 provides corresponding schema definitions.
 
-### Sink
+### Input
 
-A media consuming unit associated with an IS-04 Receiver. Sink is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
+An input of a Sender. Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
alabou: Should it says also "A media consuming unit associated with IS-04 Sender."

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny:
```suggestion
            description: Returned when the requested Input ID does not exist.
```

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny:
```suggestion
            description: Returned when the requested Output ID does not exist.
```

docs/Behaviour - Server Side.md
```suggestion
The `/inputs` endpoint returns a list of IDs corresponding to Inputs associated to the specific Sender. If this list changes then the associated [IS-04][IS-04] Sender MUST update its version (in registered mode this MUST update the registered resource).
```

docs/Data Model.md
```suggestion
The IS-04 hierarchy is preserved with the exception of the addition of Sender can identify an association with 0 or more Input resources and Receiver can identify an association with 0 or more Output resources.
```

docs/Data Model.md
```suggestion
- At the presence of a physical source or another producing unit
```
What about this?

docs/Data Model.md
N-Nagorny: You're right. It says IDs own other IDs which confused me.
```suggestion
- Device ID
```

docs/Data Model.md
```suggestion
- Device ID
```

docs/Overview.md
N-Nagorny:
```suggestion
A media producing unit associated with an IS-04 Sender. Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
Yes, but maybe _producing_.

alabou:
It depends on what we are modeling. A monitor is a media consuming unit BUT
the circuit driving the HDMI output connector is a media producing unit ...
The EDID is owned by the media consuming unit. The original intent of IS-11
as I understand it was to describe a monitor so media consuming unit. If
the input is describing for example a bluray device then a media producing
unit BUT the EDID is not owned by the bluray player but the HDMI capture
circuit which is a media consuming unit ... This is where the ambiguity
comes from. It looks strange to describe the bluray player as owning an
EDID and it may confuse people to see that both outputs and inputs are
media consuming units .... Not sure what the best solution is ...

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -36,6 +36,17 @@ mediaType: application/json
             type: !include schemas/sender-base.json
         404:
           description: Returned when the requested Sender ID does not exist.
+    /inputs:
+      get:
+        description: List of Inputs connected to the Sender.
```
cristian-recoseanu:
```suggestion
        description: List of Inputs associated with the Sender.
```

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -104,47 +115,95 @@ mediaType: application/json
             type: !include schemas/receiver-base.json
         404:
           description: Returned when the requested Receiver ID does not exist.
-    /sinks:
+    /outputs:
       get:
         responses:
           200:
-            description: List of Sinks connected to the Receiver.
+            description: List of Outputs connected to the Receiver.
```
cristian-recoseanu:
```suggestion
            description: List of Outputs associated with the Receiver.
```

README.md
```diff
@@ -9,8 +9,8 @@ This repository holds the source for this Specification, part of the family of [
 
 ### What does it do?
 
-- Provides infomation about physical devices called Sinks associated with Receivers
-- Allows to configure media parameters of Senders basing on information about Sinks
+- Introduces Inputs associated with Sender and Outputs associated with Receivers
+- Allows to configure media parameters of Senders through its Inputs basing on information about Outputs of Receivers
```
cristian-recoseanu:
```suggestion
- Allows the configuration of media parameters of Senders and Inputs using information coming from Receivers and Outputs.
```

docs/Behaviour - Server Side.md
```diff
@@ -1,18 +1,20 @@
...
+The Sink Metadata Processing API provides a mechanism to change the settings associated with logical Inputs, Sources, Flows and Senders. Another use case is obtaining information about Outputs associated with Receivers.
 
 ## Senders
 
-The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](Overview.md#media-profile) to a Sender and let the underlying implementation to choose exactly which format from the set it will use. A client can pass a single Media Profile to specify the exact format.
+The `/inputs` endpoint returns a list of IDs corresponding to Inputs associated to the specific Sender. If this list changes then the associated [IS-04][IS-04] Sender MUST update its version (in registered mode this MUST update the registered resource).
+
+The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](Overview.md#media-profile) to a Sender with subsequent passing to Inputs. The Inputs let the Source and Flow to choose exactly which format from the set they will use. A client can pass a single Media Profile to specify the exact format.
```
cristian-recoseanu:
```suggestion
The `/media-profiles` endpoint allows the client to pass an array of [Media Profiles](Overview.md#media-profile) to a Sender. The sender can then make these restrictions visible to its associated Inputs. The Inputs let the Source and Flow to choose exactly which format from the set they will use. A client can pass a single Media Profile to specify the exact format.
```

docs/Behaviour - Server Side.md
```diff
@@ -23,26 +25,30 @@ When a Sender is in this state, ie. responding with a `205 Reset Content`, this
 ### Request Methods
 
 - A `GET` request returns the last successfully applied Media Profiles array.
-- A `PUT` request MUST be validated and if the Device supports the proposed media profile, it initiates an attempt to reconfigure the Device and update corresponding Flow, Source, and/or Sender properties. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.
+- A `PUT` request MUST be validated and if the Sender supports the proposed Media Profile, the Inputs are started to be reconfigured which causes updating of the corresponding Flow, Source, and/or Sender properties. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.
```
cristian-recoseanu: I am not sure we have any properties which can be observed changing on Inputs. I would take them out and just mentioned that if the Sender supports the proposed Media Profiles, the corresponding Flow, Slow and/or Sender properties might be updated.

docs/Data Model.md
```diff
@@ -1,6 +1,6 @@
 # Data Model: Identifier Mapping
 
-[IS-04][IS-04] uses a logical data model which is extended in this specification with the addition of the Sink resources.
+[IS-04][IS-04] uses a logical data model which is extended in this specification with the addition of the Inputs and Outputs resources.
```
cristian-recoseanu: I would take this sentence out. Inputs and Outputs are data models we are adding to IS-11. They do not exist in IS-04. We are not extending the IS-04 model.
What we could say is that IS-11 builds upon existing data models in IS-04 and adds two additional resources in the form of Inputs and Outputs.

docs/Data Model.md
```diff
@@ -9,11 +9,18 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 ![Extended IS-04 Model Diagram](../docs/images/is-11-data-model.png)
 
-The IS-04 heirarchy is preserved with the exception of the addition of a Receiver can identify an association with 0 or more Sink resources.
+The IS-04 hierarchy is preserved with the exception of the addition of Sender can identify an association with 0 or more Input resources and Receiver can identify an association with 0 or more Output resources.
```
cristian-recoseanu:
```suggestion
The IS-04 hierarchy is preserved. Additionally, Senders can be associated with 0 or more Input resources and Receiver can be associated with 0 or more Output resources.
```

docs/Data Model.md
cristian-recoseanu: Are you trying to say that an IS-11 sender must have at the very least an associated Input?

docs/Data Model.md
cristian-recoseanu: Well I agree with the change in that the Output is Owned by the whole device but it's association may be to a single receiver or multiple receivers.

docs/Interoperability.md
```diff
@@ -53,6 +53,6 @@ The UUIDs used to advertise Senders and Receivers in the Sink Metadata Processin
...
+In order to prevent unnecessary polling of the Sink Metadata Processing API, changes to Inputs and Outputs are signalled via the IS-04 versioning mechanism. When the Input/Output parameters of a Sender/Receiver are modified, or when set of Inputs/Outputs associated with the Sender/Receiver is changed, the `version` attribute of the relevant IS-04 Sender/Receiver MUST be incremented.
 
-If `/media-profiles` PUT or DELETE operation applied to a Sender causes change of a Source or a Flow, their versions MUST be incremented as well. If such change causes subsequent change of Connection API `transport_file`, version of the relevant IS-04 Sender MUST be incremented as IS-05 dictates.
+If `/media-profiles` PUT or DELETE operation applied to a Sender causes change of Inputs, Source or Flow of the Sender, their versions MUST be incremented as well. If such change causes subsequent change of Connection API `transport_file`, version of the relevant IS-04 Sender MUST be incremented as IS-05 dictates.
```
cristian-recoseanu:
```suggestion
If `/media-profiles` PUT or DELETE operation applied to a Sender causes a change of Inputs, Source or Flow, their versions MUST be incremented as well. If such a change causes subsequent changes to the Connection API `transport_file`, version of the relevant IS-04 Sender MUST be incremented as IS-05 dictates.
```

docs/Overview.md
cristian-recoseanu:
```suggestion
A media producing unit which may be associated with one or multiple IS-04 Senders. Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```

docs/Overview.md
```diff
@@ -31,13 +31,17 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
...
+
+### Output
+
+A media consuming unit associated with an IS-04 Receiver. Output is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
cristian-recoseanu:
```suggestion
A media consuming unit which may be associated with one or multiple IS-04 Receivers. Output is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```

docs/Data Model.md
N-Nagorny: Yes, it was in my head at that moment. If we create Inputs without EDID for SDI sources, what is the use case for Sender without Inputs? The same can be asked about Receiver without Outputs.

docs/Behaviour - Server Side.md
N-Nagorny: If we say Input owns EDID, then it might be updated as well.

docs/Data Model.md
```suggestion
```

docs/Overview.md
N-Nagorny:
```suggestion
The purpose of AMWA IS-11 is to provide a mechanism by which to configure Sources, Flows and Senders using information from Receivers. Receiver provides Receiver Capabilities and exposes its Outputs each of which may have EDID. Sender has Media Profiles, which restrict media formats allowed for sending, and Inputs each of which may have an effective EDID to present to its upstream media producing unit.
```

docs/Behaviour - Server Side.md
alabou: What about this?

- A `PUT` request MUST be validated and if the Sender supports the proposed Media Profiles, the associated Inputs may be reconfigured to adhere to the Media Profiles constraints, which may eventually cause the update of the corresponding Flows, Sources, and Sender. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.

docs/Behaviour - Server Side.md
```diff
@@ -23,26 +25,30 @@ When a Sender is in this state, ie. responding with a `205 Reset Content`, this
...
+The `/outputs` endpoint returns a list of IDs corresponding to Outputs associated to the specific Receiver. If this list changes then the underlying [IS-04][IS-04] Receiver MUST update its version (in registered mode this MUST update the registered resource).
 
 If Receiver becomes unable to handle Flow, it is REQUIRED to take the following actions:
 - MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
 - MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
 
-## Sinks
+## Inputs
+
+The `/edid` endpoint allows a client to download the effective EDID if it exists. If the EDID information for the Input changes and it is associated with any Senders, then all of the Senders in question MUST update their versions (in registered mode this MUST update the registered resources).
```
alabou, line 47: Why requiring Sender version change if EDID changes? If the EDID cause a change of Source or Flow the Sender will be updated, if the EDID changes without causing a change in Source or Flow then why update the version of the Sender? On the Input side the EDID changes through IS-11 so the controller knows about it.

N-Nagorny: Controller knows about Input change when it PUTs EDID to this Input. If Controller PUTs Media Profiles to a Sender, the Input's EDID may or not change, and Sender's version update indicates it even if Source and Flow stays the same. If the Input is shared between two Senders, it helps Controller to keep them up-to-date.

Another reason is the multiple NMOS APIs clients case.

alabou: Isn't it that PUT media_profile will bump by itself the version of the Sender?

N-Nagorny: Right, it is. I assume, providing effective EDID binary to HDMI source is not immediate. Controller PUTs Media Profiles, the Sender responds and updates own version. Then the HDMI input sets HPD low, waits few hundreds of msecs, sets HPD high, the HDMI source reads the fresh effective EDID binary, the Input updates its effective EDID and updates own version and version of the Sender.

If effective EDID models not the last EDID read by HDMI source but the last EDID generated by Sender, we may treat Media Profiles change and effective EDID change as immediate. But I think bumping version of the Sender after effective EDID update is still useful for Inputs associated with multiple Senders and NMOS APIs clients other than the Controller.

alabou: I don't think we want/need to model the HPD to EDID read by the source part of the system ... Eventually the source will get the effective EDID and behave accordingly, possibly changing the signal it transmits ... I would not want to bump the version of a Sender to indicate that after HPD a source has read the effective EDID ...

docs/Behaviour - Server Side.md
```suggestion
- A `PUT` request MUST be validated and if the Sender supports the proposed Media Profiles, the associated Inputs may be reconfigured to adhere to the Media Profiles constraints, which may eventually cause the update of the corresponding Flows, Sources, and Sender. If successful, the operation MUST return the accepted Media Profiles array, otherwise an error MUST be returned and nothing changed.
```

docs/Overview.md
```diff
@@ -31,13 +31,17 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
 
 The terms 'Node', 'Device', 'Sender' and 'Receiver' are used extensively in this documenation set. The [NMOS Technical Overview](https://specs.amwa.tv/nmos/main/docs/2.0._Technical_Overview.html) provides an outline of these terms, and IS-04 provides corresponding schema definitions.
 
-### Sink
+### Input
 
-A media consuming unit associated with an IS-04 Receiver. Sink is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
+A media producing unit which may be associated with one or multiple IS-04 Senders. Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
N-Nagorny:
```suggestion
An Input is a unit consuming media data for providing it to Sender(s). Input may present metadata about media consuming capabilities to its upstream counterpart. Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```

docs/Overview.md
```diff
@@ -31,13 +31,17 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
...
+
+### Output
+
+A media consuming unit which may be associated with one or multiple IS-04 Receivers. Output is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
N-Nagorny:
```suggestion
An Output is a unit producing media data provided by Receiver(s). Output may consume metadata about the media consuming capabilities from its downstream counterpart. Output is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```

docs/Data Model.md
N-Nagorny: I plan to manually commit changes to Identity Hierarchy section if we agree Sender/Receiver<-->Input/Output relationship is `1..*`.

Regarding Input/Output Representation sections, the only obvious circumstance to create them for me is the presence of a physical source/sink device which reads/provides EDID. I guess, such language leaves any other appropriate cases open for vendors.

alabou: A Sender is associated to 1..* Inputs, an Input is associated to 1..* Senders. Idem for Receivers and Outputs. I'm ok with that.

docs/Behaviour - Server Side.md
N-Nagorny: It makes sense not to link Input behaviour with physical source. Does it make sense to keep bumping version of the Sender after effective EDID update for Inputs with multiple Senders and multiple NMOS clients or there are better alternatives?

docs/Data Model.md
N-Nagorny: I made a commit with bidirectional `1..*` relationship and updated Representation. Marked as resolved.

docs/Behaviour - Server Side.md
N-Nagorny: 2020/10/25 activity group call: PUT for /media-profiles and /edid doesn't wait for passing EDID binary to a physical source; /edid change bumps Sender version to let all the Senders associated with the Input be tracked by Controller. Marked as resolved.

PR #67, opened 2021-10-27
Describe a case of non-IS-11 Senders/Receivers

N-Nagorny:


docs/Data Model.md
```diff
@@ -8,7 +8,7 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 ![Extended IS-04 Model Diagram](../docs/images/is-11-data-model.png)
 
-The IS-04 hierarchy is preserved. Additionally, Senders can be associated with 1 or more Input resources and Receiver can be associated with 1 or more Output resources.
+The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 SHOULD be associated with 1 or more Input resources and Receivers managed with IS-11 SHOULD be associated with 1 or more Output resources. Senders and Receivers not managed with IS-11 MAY be associated with Inputs and Outputs respectively.
```
alabou: I think that without IS-11 there is no way to associate to inputs/outputs so the last sentence should not be there.

```suggestion
The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 SHOULD be associated with 1 or more Input resources and Receivers managed with IS-11 SHOULD be associated with 1 or more Output resources.
```

PR #71, opened 2021-11-15
Arbitrary Base EDID setter

N-Nagorny:


APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -208,21 +208,96 @@ mediaType: application/json
...
+              description: A temporary redirect to the Base EDID file.
+            404:
+              description: >
+                Returned when the requested Input ID doesn't exist.
+              body:
+                type: !include schemas/error.json
+        put:
+          description: Replace Base EDID with new one
+          responses:
+            202:
```
alabou: I think that 200 is more appropriate. The base EDID has been accepted by the device and becomes the current EDID. Are you returning 202 because the new base EDID will not be used until new constraints are PUT?

N-Nagorny: It makes sense to me to use Base EDID right after accepting it.

https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.1 says "Aside from responses to CONNECT, a 200 response always has a payload, though an origin server MAY generate a payload body of zero length". I don't know if it makes sense to put the accepted Base EDID binary to the payload so I chose 202 Accepted with no body.

docs/Interoperability.md
```diff
@@ -11,7 +11,7 @@ IS-05 dictates that IS-04 APIs SHOULD be operating at version 1.2 or greater in
 
 Any references to terms and JSON Schemas defined in BCP-004-01 relate to version 1.0.0.
 
-Receivers listed under IS-11 `/receivers` endpoint MUST use BCP-005-01 to express their Receiver Capabilities.
+Receivers, managed with IS-11 and associated with Outputs which support EDID, SHOULD use BCP-005-01 to express Receiver Capabilities came from the EDID(s).
```
N-Nagorny:
```suggestion
Receivers, managed with IS-11 and associated with Outputs which support EDID, SHOULD use BCP-005-01 to express Receiver Capabilities came from the EDID(s) and BCP-004-01 to express their own Receiver Capabilities. The inherent Receiver Capabilities prevail over the EDID associated Capabilities. An EDID can further restrict inherent Receiver Capabilities but cannot expand them.
```

APIs/SinkMetadataProcessingAPI.raml
alabou: Why not 204 then? But there is another issue: if the input does not support and EDID (i.e. SDI) then a GET return 204 ... I think a PUT should also be able to indicate that the input does not support an EDID and should also return 204 ... which becomes bad because success without body response is similar ... Would a PUT returning an empty body with 200 be better ????

N-Nagorny: I replaced 202 with 204 so that if an Input/Output doesn't support/provide EDID, then all GETs under `/edid` return `204 No Content`, `PUT /edid/base` returns `405 Method Not Allowed` and `edid_support` is `false` in `/properties`. If an Input supports EDID but doesn't support setting arbitrary Base EDID, then `PUT /edid/base` returns `405 Method Not Allowed`. I assume this case is rare enough to signalize it with an additional boolean.

docs/Behaviour - Server Side.md
```diff
@@ -31,23 +31,27 @@ If an active Receiver becomes unable to handle Flow, it is REQUIRED to take the
...
+The `/properties` endpoint shows properties of the Input. In particular, it MAY contain information parsed from Base EDID and Effective EDID if the Input supports EDID and the Node is capable to parse EDID binaries.
 
 ### Base EDID
 
-`GET /edid/base` and `PUT /edid/base` operations allow a client to download and replace the Base EDID respectively if it exists. If the Base EDID for the Input changes and it is associated with any Senders, then all of the Senders in question MUST update their versions (in registered mode this MUST update the registered resources).
+The initial state of Base EDID is such an EDID which this Input presents to the upstream counterpart if Active Constraints are empty.
+
+`GET /edid/base` and `PUT /edid/base` operations allow a client to download and replace the Base EDID respectively if it exists. `DELETE /edid/base` resets Base EDID to the initial state. If the Base EDID for the Input changes and it is associated with any Senders, then all of the Senders in question MUST update their versions (in registered mode this MUST update the registered resources).
```
alabou: There will be no way of knowing of a PUT has been performed on /edid/base if we return the manufacturer default base EDID after a delete to /edid/base. Is it appropriate according to REST rules? Another approach to get the manufacturer default EDID would be a GET on /edid/effective with empty /constraints/active

docs/Interoperability.md
```diff
@@ -11,7 +11,7 @@ IS-05 dictates that IS-04 APIs SHOULD be operating at version 1.2 or greater in
 
 Any references to terms and JSON Schemas defined in BCP-004-01 relate to version 1.0.0.
 
-Receivers, managed with IS-11 and associated with Outputs which support EDID, SHOULD use BCP-005-01 to express Receiver Capabilities came from the EDID(s).
+Receivers, managed with IS-11 and associated with Outputs which support EDID, SHOULD use BCP-005-01 to express Receiver Capabilities came from the EDID(s) and BCP-004-01 to express their own Receiver Capabilities. The inherent Receiver Capabilities prevail over the EDID associated Capabilities. An EDID can further restrict inherent Receiver Capabilities but cannot expand them.
```
alabou: missing "that" in "Capabilities that came from the EDID(s)" I think ...

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -59,26 +59,26 @@ mediaType: application/json
               type: !include schemas/constraints-base.json
       /active:
         get:
-          description: Show the last applied Constraints.
+          description: Show the Active Constraints.
```
alabou, line 66: Missing 404 on /constraints above ...

docs/Interoperability.md
N-Nagorny:
```suggestion
Receivers, managed with IS-11 and associated with Outputs which support EDID, SHOULD use BCP-005-01 to express Receiver Capabilities that came from the EDID(s) and BCP-004-01 to express their own Receiver Capabilities. The inherent Receiver Capabilities prevail over the EDID associated Capabilities. An EDID can further restrict inherent Receiver Capabilities but cannot expand them.
```

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny: I fixed that. Thanks!

docs/Behaviour - Server Side.md
N-Nagorny: You're perfectly right. I replaced default Base EDID with No Content.

docs/Behaviour - Server Side.md
```diff
@@ -31,19 +31,17 @@ If an active Receiver becomes unable to handle Flow, it is REQUIRED to take the
...
 
-The initial state of Base EDID is such an EDID which this Input presents to the upstream counterpart if Active Constraints are empty.
-
-`GET /edid/base` and `PUT /edid/base` operations allow a client to download and replace the Base EDID respectively if it exists. `DELETE /edid/base` resets Base EDID to the initial state. If the Base EDID for the Input changes and it is associated with any Senders, then all of the Senders in question MUST update their versions (in registered mode this MUST update the registered resources).
+There is no Base EDID at the initial state. If Base EDID for an Input changes, then all Senders associated with this Input MUST update their versions (in registered mode this MUST update the registered resources).
 
 ### Effective EDID
 
-Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+Effective EDID is such EDID that the baseband signal from the Input can be transmitted by the Senders associated with this Input without breaking their Active Constraints.
```
alabou: I<m not sure I understand what this mean. I liked the original "Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input"

N-Nagorny: I reverted it and added language about what happens if there's no Base EDID set needed for this combination.

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny: 2021/11/22 activity group call: marked as resolved.

docs/Behaviour - Server Side.md
N-Nagorny: 2021/11/22 activity group call: marked as resolved.

PR #74, opened 2021-11-30
Add Receiver Status

N-Nagorny:


docs/Behaviour - Server Side.md
```diff
@@ -16,26 +16,30 @@ Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the
...
 - `Constrained` when this Sender, the Flow and the `/transportfile` (if used) associated with this Sender and the Source associated with this Flow satisfy the Active Constraints.
 - `Active Constraints Violation` when this Sender, the Flow or the `/transportfile` (if used) associated with this Sender or the Source associated with this Flow does not satisfy the Active Constraints. When an inactive Sender is in this state, it MUST NOT allow [IS-05][IS-05] activations.
 - `No Signal` when there is no signal from Inputs associated with this Sender.
 - `Awaiting Signal` when Active Constraints was just PUT and the signal from Inputs associated with this Sender is not stable yet. This is a transitional state until one of the previous ones can be established.
 
+## State of Receiver
+
+A Receiver managed with IS-11 has the following states:
+- `No Transport File` when there's no active `transport_file`.
+- `OK` when this Receiver's active `transport_file` does not violate Receiver Capabilities.
```
alabou, line 30: Can it be that if there is no transport file it cannot violate the capabilities and then OK is enough?

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -57,8 +57,8 @@ mediaType: application/json
           200:
             description: Returned when success.
             body:
-              example: !include ../examples/status-get-200.json
-              type: !include schemas/status.json
+              example: !include ../examples/sender-status-get-200.json
```
alabou, line 60: Missing status to sender-base.json

APIs/SinkMetadataProcessingAPI.raml
```diff
@@ -180,6 +180,19 @@ mediaType: application/json
           description: Returned when the requested Receiver ID does not exist.
           body:
             type: !include schemas/error.json
+    /status:
```
alabou, line 183: Missing status to receiver-base.json

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny: I fixed that. Thank you!

APIs/SinkMetadataProcessingAPI.raml
N-Nagorny: Fixed.

docs/Behaviour - Server Side.md
N-Nagorny: I would say if there is no transport file, the Receiver doesn't know whether the Flow violates Receiver Capabilities which is a separate status.

alabou: Is it really possible to connect a receiver without a transport_file? If YES then I'm ok with this status code.

N-Nagorny: IS-05 gives such option. I guess, there are codecs which embed metadata about media in packet flow and I tried to mark transport file usage as optional everywhere in IS-11.

alabou (approved):

PR #77, opened 2021-12-17
Fix typo Outpus to Outputs

garethsb:
And add a lot of indefinite or definite articles, hope that's OK!

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -9,7 +9,7 @@
 
 ## Introduction
 
-The purpose of AMWA IS-11 is to provide a mechanism by which to configure Sources, Flows and Senders using information from Receivers. Receiver provides Receiver Capabilities and exposes its Outputs each of which may hold EDID from its downstream counterpart. Sender has Active Constraints, which restrict media formats allowed for sending, and Inputs each of which may have an Effective EDID to present to its upstream media producing unit.
+The purpose of AMWA IS-11 is to provide a mechanism by which to configure Sources, Flows and Senders using information from Receivers. A Receiver provides Receiver Capabilities and exposes its Outputs, each of which may hold EDID from its downstream counterpart. A Sender has Active Constraints, which restrict media formats allowed for sending, and Inputs, each of which may have an Effective EDID to present to its upstream media producing unit.
 
 The Specification includes:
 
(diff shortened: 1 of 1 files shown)


docs/Overview.md
```diff
@@ -35,37 +35,37 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
...
 
 ### Senders
 
-Each Sender has `inputs` resource, which lists identifiers of all Inputs associated with this Sender, `constraints` base resource in charge of restrictions for the Flow associated with this Sender and `status` resource showing Sender's state in terms of IS-11.
+Each Sender has an `inputs` resource, which lists identifiers of all Inputs associated with this Sender, a `constraints` base resource in charge of restrictions for the Flow associated with this Sender and a `status` resource showing the Sender's state in terms of IS-11.
 
 #### Constraints
 
-There is `active` resource which holds Active Constraints and `supported` resource which lists all the Parameter Constraints supported by this Sender.
+There is an `active` resource which holds Active Constraints and a `supported` resource which lists all the Constraint Sets supported by this Sender.
```
N-Nagorny: Thank you Gareth! The only thing is that a Sender lists supported Parameter Constraints to show which of them make sense for the Sender and it can try to adhere to them.

```suggestion
There is an `active` resource which holds Active Constraints and a `supported` resource which lists all the Parameter Constraints supported by this Sender.
```

garethsb: Ah, sorry, yes!!

garethsb:
@N-Nagorny two notes...

We need to confirm that `draft-07` is acceptable to the community as I _think_ all current specs are `draft-04` so far. I wish the JSON Schema group would get on with it and publish a final spec and we could think about moving to that!

I see the schema has a `contains` and `minContains`. Is that intended to make that whole list required for all implementations?

N-Nagorny:
@garethsb, thanks, a quick check shows that the schemas are `draft-04` compatible so I'll just create a PR replacing `$schema` value.

`contains` and `minContains` are added intentionally, otherwise there may be a situation when a sending Device advertises IS-11 support but it doesn't allow to change basic parameters.

N-Nagorny:
Merged on the meeting on January 24, 2022. Thank you, @garethsb!

PR #82, opened 2022-01-12
Memorable Base URI

### LINKED ISSUE #78: Spec name and API short name
Flow Compatibility Management is a good name.
_/x-nmos/fcm_ is somehow not as memorable.

Other NMOS APIs, even ones with long-winded names like Audio Channel Mapping so far decided not to use initialisms, so I'm just wondering if _/x-nmos/flowcompatibility_ or something would be better?

Hm, it is _Flow_ compatibility we're talking about here, not _Stream_ compatibility, is it?? 🤔

N-Nagorny:
Resolves #78.


alabou (approved):
Ok for me.

alabou:
As a generic question: why it is not the IS number like "is-11" that is used in NMOS to identify an API such that instead of "fc-ctrl" or "flowcompatibility" it would simply be "is-11" ?

N-Nagorny:
Merged on the group meeting on January 24, 2022.

PR #83, opened 2022-01-25
Split version increments into related and unrelated to the API

N-Nagorny:


docs/Interoperability.md
```diff
@@ -55,6 +55,6 @@ The UUIDs used to advertise Senders and Receivers in the Flow Compatibility Mana
 
 ## Version Increments
 
-In order to prevent unnecessary polling of the Flow Compatibility Management API, changes to Inputs and Outputs are signalled via the IS-04 versioning mechanism. When the Input/Output parameters of a Sender/Receiver are modified, when set of Inputs/Outputs associated with the Sender/Receiver is changed, or when Status of Sender/Receiver is updated, then the `version` attribute of the relevant IS-04 Sender/Receiver MUST be incremented.
+In order to prevent unnecessary polling of the Flow Compatibility Management API, changes to it are signalled via the IS-04 versioning mechanism. E.g. when the Input/Output parameters of a Sender/Receiver are modified, when set of Inputs/Outputs associated with the Sender/Receiver is changed, or when Status of Sender/Receiver is updated, then the `version` attribute of the relevant IS-04 Sender/Receiver MUST be incremented.
```
alabou: I would not  use "changes to it" ... but simply "changes"

I would add "when state of Inputs/Outputs associated with the Sender/Receiver changes"

I would rephrase "In order to prevent unnecessary polling of the Flow Compatibility Management API, changes are signaled via the IS-04 versioning mechanism of the associated Sender and Receiver objects"

docs/Interoperability.md
```diff
@@ -55,6 +55,6 @@ The UUIDs used to advertise Senders and Receivers in the Flow Compatibility Mana
...
 
-If `PUT /constraints/active`, `DELETE /constraints/active` or `PUT /edid/base` operation applied to a Sender causes a change of Inputs, Source or Flow, their versions MUST be incremented as well. If such a change causes changes to the Connection API `/transportfile` resource, version of the relevant IS-04 Sender MUST be incremented.
+If any resource such as Source, Flow, Sender etc. changes over its lifetime, then its `version` attribute MUST be incremented as well.
```
alabou: I would rephrase "If the state of any resource such as Source, Flow, Sender, Receiver, etc. changes over its lifetime, then its `version` attribute MUST be incremented as well."

docs/Interoperability.md
N-Nagorny: Thanks, it sounds better.

```suggestion
In order to prevent unnecessary polling of the Flow Compatibility Management API, changes are signaled via the IS-04 versioning mechanism of the associated Sender and Receiver resources. E.g. when the Input/Output parameters of a Sender/Receiver are modified, when set of Inputs/Outputs associated with the Sender/Receiver is changed, or when Status of Sender/Receiver is updated, then the `version` attribute of the relevant IS-04 Sender/Receiver MUST be incremented.
```

docs/Interoperability.md
N-Nagorny: It seems more confusing to me because we use term "state" for Sender/Receiver. What about "properties"?

```suggestion
If properties of any resource such as Source, Flow, Sender, Receiver, etc. change over its lifetime, then its `version` attribute MUST be incremented as well.
```

docs/Interoperability.md
alabou: Maybe "properties" instead of "parameters" ... ?

docs/Interoperability.md
alabou: ok

docs/Interoperability.md
N-Nagorny:
```suggestion
In order to prevent unnecessary polling of the Flow Compatibility Management API, changes are signaled via the IS-04 versioning mechanism of the associated Sender and Receiver resources. E.g. when properties of any Input/Output associated with the Sender/Receiver are modified, when set of Inputs/Outputs associated with the Sender/Receiver is changed, or when Status of Sender/Receiver is updated, then the `version` attribute of the relevant IS-04 Sender/Receiver MUST be incremented.
```

PR #84, opened 2022-01-25
Preventing violations

N-Nagorny:


APIs/FlowCompatibilityManagementAPI.raml
```diff
@@ -89,19 +89,24 @@ mediaType: application/json
               body:
                 type: !include schemas/error.json
         put:
-          description: Replace Active Constraints with other ones.
+          description: >
+            Validates proposed Constraints against JSON Schema of request body
+            and Supported Parameter Constraints.
```
alabou: Seems to be missing that it replace constraints .... "and upon success replace active constraints with the new ones."

APIs/FlowCompatibilityManagementAPI.raml
```diff
@@ -89,19 +89,24 @@ mediaType: application/json
...
           body:
             example: !include ../examples/constraints-active-get-200.json
             type: !include schemas/constraints_active.json
           responses:
             200:
-              description: Returned when success.
+              description: >
+                Returned when this Sender is capable to adhere at least one
```
alabou: This seems to be misleading ... the Sender accept the constraints if they are valid and included in the supported constraints ... and possibly if the extents of the constraints are supported by the Sender ... BUT it does not mean the Sender will be able to adhere to the constraints ... it all depends on the effective input signal.

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
-- `DELETE` request MUST clear Active Constraints.
+Active Constraints are satisfied when any of its Constraint Sets are satisfied in terms of [BCP-004-01][BCP-004-01].
 
 ## State of Sender
 
 A Sender managed with IS-11 has the following states:
 - `Unconstrained` when Active Constraints of this Sender is an empty `constraint_sets` array.
-- `Constrained` when this Sender, the Flow and the `/transportfile` (if used) associated with this Sender and the Source associated with this Flow satisfy the Active Constraints.
-- `Active Constraints Violation` when this Sender, the Flow or the `/transportfile` (if used) associated with this Sender or the Source associated with this Flow does not satisfy the Active Constraints. When an inactive Sender is in this state, it MUST NOT allow [IS-05][IS-05] activations.
+- `Constrained` when the Active Constraints are satisfied.
```
alabou, line 16: "when at least one of the Active Constraints is satisfied."

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
+- `Active Constraints Violation` when at least one Parameter Constraint of each Constraint Set is not satisfied.
```
alabou: "when none of the Active Constraints is satisfied."

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
 - `No Signal` when there is no signal from Inputs associated with this Sender.
-- `Awaiting Signal` when Active Constraints was just PUT and the signal from Inputs associated with this Sender is not stable yet. This is a transitional state until one of the previous ones can be established.
+- `Awaiting Signal` when Inputs associated with this Sender were reconfigured recently and the signal is not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update such associated resources as Flow and Source while this Sender is in this state.
```
alabou: "when the signal from Inputs associated to this Sender is transitioning and the signal is not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
 Non-empty `constraint_sets` array MUST be treated as unordered unless at least one Constraint Set has `urn:x-nmos:cap:meta:preference` attribute.
 
-Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the Connection API `/transportfile` resource (if used) associated with this Sender and the Source associated with this Flow MUST satisfy the Active Constraints when the Sender is active.
-
-### Request Methods
-
-- `GET` request returns the last successfully applied Active Constraints.
-- `PUT` request MUST be validated and if the Sender supports Parameter Constraints used in the proposed Active Constraints and is capable to adhere at least one of the Constraint Sets, then the associated Inputs may be reconfigured, which may eventually cause the update of the corresponding Flows, Sources, `/transportfile` and Sender. If successful, the operation MUST return the accepted Active Constraints, otherwise an error MUST be returned and nothing changed.
-- `DELETE` request MUST clear Active Constraints.
+Active Constraints are satisfied when any of its Constraint Sets are satisfied in terms of [BCP-004-01][BCP-004-01].
```
alabou, line 9: I would keep this to make sure a Sender cannot advertise through  IS-04 a Flow or through IS-05 a transport file that violate the active constraints. If we simply state that a Sender cannot transmit to the network it keeps the possibility for a sender to advertise an illegal Flow and SDP transport file ...

"Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the Connection API `/transportfile` resource (if used) associated with this Sender and the Source associated with this Flow MUST satisfy the Active Constraints when the Sender is active."

APIs/FlowCompatibilityManagementAPI.raml
N-Nagorny:
```suggestion
            and Supported Parameter Constraints and upon success
            replaces Active Constraints with the new ones.
```

APIs/FlowCompatibilityManagementAPI.raml
N-Nagorny: Maybe we need better wording here. My understanding is that ability/capability to adhere means the Sender knows these values and will try and doesn't guarantee that Sender will adhere after the PUT.

docs/Behaviour - Server Side.md
N-Nagorny: Great wording.

```suggestion
- `Awaiting Signal` when the signal from Inputs associated to this Sender is transitioning and the signal is not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
```

docs/Behaviour - Server Side.md
N-Nagorny: Active Constraints is an object with a single property "constraint_sets". I added "Active Constraints are satisfied when any of its Constraint Sets are satisfied in terms of [BCP-004-01][BCP-004-01]" to define what Active Constraints satisfaction means. If any properties are added to Active Constraints, it will be possible to change this single line instead of editing all the places referring to the satisfaction.

docs/Behaviour - Server Side.md
N-Nagorny: Taking into account the comment above, should I add a definition of Active Constraints dissatisfaction and refer to it here?

docs/Behaviour - Server Side.md
N-Nagorny: I don't quite understand the point. A Sender has some Flow and transport file. Each time they change or Active Constraints change, the Node validates the Flow and the SDP against the Active Constraints. If validation fails, the Node prohibits transmitting with an illegal Flow and SDP. I don't understand, how we can avoid advertising an illegal Flow and SDP at all.

The removed paragraph doesn't save us from it as well because of "when the Sender is active" so I think "Preventing restrictions violation" in its current state covers this removed paragraph.

APIs/FlowCompatibilityManagementAPI.raml
alabou: ok

docs/Behaviour - Server Side.md
alabou: ok

docs/Behaviour - Server Side.md
alabou: Yes, would be more uniform ...

docs/Behaviour - Server Side.md
alabou: ok

docs/Behaviour - Server Side.md
alabou: I would not want a Sender to be allowed to publish an associated Flow/Source that violate the active constraints. Just not transmitting to the network it not enough. This paragraph provide more control over such behavior. This is why I think we must keep it. What would be the reason to remove it?

N-Nagorny, line 9:
```suggestion
Active Constraints are satisfied when any of its Constraint Sets are satisfied in terms of [BCP-004-01][BCP-004-01].
Active Constraints are violated when at least one Parameter Constraint of each Constraint Set is not satisfied in terms of [BCP-004-01][BCP-004-01].
```

N-Nagorny: I removed it to have this language under "Preventing restrictions violation" and it seems to me to be covered there. 

> Once a Sender accepts proposed Active Constraints, <...> Flow <...> MUST satisfy the Active Constraints **when the Sender is active**.

For me it means that either the published Flow is illegal and Sender can't transmit it or it's legal and Sender can transmit it. This logic is described in "Preventing restrictions violation" now. Could you please explain how this paragraph prevents from publishing an illegal Flow?

alabou: Without this text "Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the Connection API `/transportfile` resource (if used) associated with this Sender and the Source associated with this Flow MUST satisfy the Active Constraints when the Sender is active." the only restriction imposed by IS-11 is to not transmit a stream violating the active constraints over the network. IS-11 says nothing else. This is why I think it is important to include this text to add the additional restrictions that an active sender cannot publish an illegal flow ... only an inactive sender would be allowed.

docs/Behaviour - Server Side.md
```suggestion
- `Active Constraints Violation` when Active Constraints are violated.
```

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
+- `No Transport File` when there's no active transport file.
+- `OK` when this Receiver's active transport file does not violate Receiver Capabilities.
+- `Receiver Capabilities Violation` when this Receiver's active transport file violates Receiver Capabilities.
 
 ## Preventing restrictions violation
 
-At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST NOT transmit the Flow over the network. The Sender is REQUIRED to take the following actions:
-- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
-- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
+At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST NOT transmit the Flow over the network. An inactive Sender in this state MUST NOT allow activations.
```
```suggestion
At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST become inactive. An inactive Sender in this state MUST NOT allow activations.
```

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,29 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
 
-At any time if State of an active Receiver becomes `Receiver Capabilities Violation`, the Receiver MUST stop receiving the Flow by taking the following actions:
-- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
-- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
+At any time if State of an active Receiver becomes `Receiver Capabilities Violation`, the Receiver SHOULD stop receiving the Flow. An inactive Receiver in this state SHOULD NOT allow activations.
```
```suggestion
At any time if State of an active Receiver becomes `Receiver Capabilities Violation`, the Receiver SHOULD become inactive. An inactive Receiver in this state SHOULD NOT allow activations.
```

APIs/FlowCompatibilityManagementAPI.raml
```diff
@@ -89,19 +89,25 @@ mediaType: application/json
...
+            replaces Active Constraints with the new ones.
           body:
             example: !include ../examples/constraints-active-get-200.json
             type: !include schemas/constraints_active.json
           responses:
             200:
-              description: Returned when success.
+              description: >
+                Returned when this Sender is capable to adhere at least one
+                of the proposed Constraint Sets.
```
N-Nagorny:
```suggestion
                Returned when this Sender supports Parameter Constraint URNs used
                and all of Parameter Constraint values in all of Constraint Sets.
```

APIs/FlowCompatibilityManagementAPI.raml
```diff
@@ -89,19 +89,25 @@ mediaType: application/json
...
               body:
                 example: !include ../examples/constraints-active-get-200.json
                 type: !include schemas/constraints_active.json
             400:
               description: >
-                Returned when the Constraints are rejected due to schema validation failure or if the Sender doesn't support Parameter Constraints used.
+                Returned when the Constraints are rejected due to schema validation failure
+                or if the Sender doesn't support Parameter Constraints used.
```
N-Nagorny:
```suggestion
                Returned when the Constraints are rejected due to schema validation failure
                or if the Sender doesn't support Parameter Constraint URNs used.
```

APIs/FlowCompatibilityManagementAPI.raml
```diff
@@ -110,7 +116,8 @@ mediaType: application/json
                 type: !include schemas/error.json
             422:
               description: >
-                Returned when the Sender isn't capable to adhere to the Constraints although the request is valid and the Sender supports all the Parameter Constraints used.
+                Returned when the Sender isn't capable to adhere to the proposed Constraints
+                although the request is valid and the Sender supports all the Parameter Constraints used.
```
N-Nagorny:
```suggestion
                Returned when the Sender isn't capable to adhere to the proposed Constraints
                although the request is valid and the Sender supports all the Parameter Constraint URNs used.
```

docs/Behaviour - Server Side.md
```diff
@@ -6,39 +6,30 @@ The initial state of Active Constraints of a Sender MUST be an empty `constraint
...
+- `No Transport File` when there's no active transport file.
+- `OK` when this Receiver's active transport file does not violate Receiver Capabilities.
+- `Receiver Capabilities Violation` when this Receiver's active transport file violates Receiver Capabilities.
 
 ## Preventing restrictions violation
 
-At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST NOT transmit the Flow over the network. The Sender is REQUIRED to take the following actions:
-- MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
-- MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
+At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST become inactive. An inactive Sender in this state MUST NOT allow activations.
```
alabou, line 30: OK

docs/Behaviour - Server Side.md
N-Nagorny: `Sender MUST NOT transmit` -> `Sender MUST become inactive` solved this issue, so I'm marking this as resolved.

APIs/FlowCompatibilityManagementAPI.raml
N-Nagorny: The wording is upgraded.

PR #85, opened 2022-01-31
Fix TOC generation

garethsb:


garethsb:
I think the other pages of the docs (apart from the _Behaviour_ intro) could also have TOCs added?

N-Nagorny:
https://specs.amwa.tv/is-05/ has TOC only on "Overview" page, although https://specs.amwa.tv/is-08/ has it on every page with more than one section. @peterbrightwell, are there any rules about this?

PR #89, opened 2022-01-31
Gareth's suggestions regarding language/formatting

### LINKED ISSUE #87: A few questions and some minor language/formatting issues
### Interoperability

> Any references to terms and JSON Schemas defined in BCP-004-01 relate to version 1.0.0.

Probably reasonable to mention the version number, but I suggest v1.0 rather than v1.0.0.

> `http://hostname/x-nmos/`...

I suggest e.g. `http://api.example.com/x-nmos/`... like other parts of the NMOS specs. (`example.com` is [reserved for examples by IANA](https://www.iana.org/domains/reserved).)

> The Flow Compatibility Management API should be

There are at least two instances of lowercase "should". If they are intended to be conformance key words they SHOULD be in uppercase. :-)

I suggest some more articles in a few places, e.g.

```diff
> when set of Inputs/Outputs associated with the Sender/Receiver is changed, or when Status of Sender/Receiver is updated
---
< when the set of Inputs/Outputs associated with the Sender/Receiver is changed, or when the Status of a Sender/Receiver is updated
```

### Data Model: Identity Mapping

> Output ID
> [...]
> SHOULD change if:
> [...]
> A different upstream counterpart is connected to the same interface

If I've understood correctly, that should be _downstream_?

### Behaviour: Server Side

> Non-empty constraint_sets array MUST be treated as unordered unless at least

"Treated" by which actor? The Sender? Or a Controller?

Also, I suggest "of equal preference" rather than "unordered".

> Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the Connection API `/transportfile` resource (if used) associated with this Sender and the Source associated with this Flow MUST satisfy the Active Constraints when the Sender is active.

I'm just checking... does "satisfy the Active Constraints" mean "satisfy at least one of the Constraint Sets in the Active Constraints"? If so, why does it help for the Sender to have multiple Constraint Sets associated with it?

In the section on different states, does the relative precedence of `Active Constraints Violation` and `No Signal` need to be indicated?

I think the use of capitalized space-separated strings for `enum` values may be inconsistent with other NMOS specs. If a human-readable string is required that should be a separate property, such as `description`.

I suggest lowercase, underscore separated, for example:

```json
"enum": {
  "unconstrained",
  "constrained",
  "active_constraints_violation",
  "no_signal",
  "awaiting_signal"
}
```

### Behaviour: Client Side

> If multiple Receivers have equal Constraint Sets, NMOS Controller MUST take only one of them. If they differ only in `urn:x-nmos:cap:meta:preference`, then one with the highest priority MUST be taken.

What does it mean for the Controller to "take" a constraint set?

N-Nagorny:
Resolves #87.


docs/Behaviour - Client Side.md
```diff
@@ -1,16 +1,16 @@
...
 
-If multiple Receivers have equal Constraint Sets, NMOS Controller MUST take only one of them. If they differ only in `urn:x-nmos:cap:meta:preference`, then one with the highest priority MUST be taken.
+If multiple Receivers have equal Constraint Sets, NMOS Controller MUST choose only one of them while making the intersection. If they differ only in `urn:x-nmos:cap:meta:preference`, then one with the highest priority MUST be taken.
```
garethsb: Now I understand this is a bit better, my question is whether the Controller has to use the preference values from the Receivers at all? The Controller may know that one Receiver is higher priority than another, for example? And in any case, the way `meta:preference` is defined, values can only be compared for the same Receiver, you can't compare between them.

APIs/schemas/receiver-status.json
```diff
@@ -9,9 +9,9 @@
     "state": {
       "type": "string",
       "enum": [
-        "No Transport File",
-        "OK",
-        "Receiver Capabilities Violation"
+        "no_transport_file",
+        "ok",
```
garethsb: Is it possible to be more descriptive than 'OK', I wonder?

docs/Behaviour - Client Side.md
```diff
@@ -1,16 +1,16 @@
 # Behaviour: Client Side
 
-## Constraints of Sender
+## Active Constraints of Sender
 
 Before making an [IS-05][IS-05] connection, NMOS Controller chooses a Sender and Receivers. Then NMOS Controller MUST `GET /constraints/supported` from the Sender, collect Receiver Capabilities from the Receivers and make such intersection of them that they can be utilized for building Constraints satisfying all the Receivers and using Parameter Constraints supported by the Sender. If the Sender supports less Parameter Constraints than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. After that NMOS Controller MUST `PUT /constraints/active` and make the IS-05 connection if the Constraints have been applied succesfully. After breaking this connection via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active`.
```
```suggestion
Before making an [IS-05][IS-05] connection, NMOS Controller chooses a Sender and Receivers. Then NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such intersection of them that they can be utilized for building Active Constraints satisfying all the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. After that NMOS Controller MUST `PUT /constraints/active` and SHOULD make the IS-05 connection if the Constraints have been applied succesfully. After breaking this connection via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active`.

Constraint Sets of Active Constraints SHOULD be the largest consensus among Receivers. For example, there are Receiver A, Receiver B, Receiver C with equal Receiver Capabilities consisting of Constraint Sets 1, 2, 3, 4, 5 and Receiver D with Constraint Sets 2, 3, 4, 5 and 6 which is incompatible with the previous ones. In this case the largest consensus is Constraint Sets 2, 3, 4, 5 which are common for all the four Receivers.

`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates Constraint Set preference between Constraints Sets of the Receiver. NMOS Controller MUST remove values of this property before `PUT /constraints/active` and MAY use these values while making a decision about the intersection.

`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Active Constraints indicates Constraint Set preference between Constraints Sets of the Active Constraints of the Sender. NMOS Controller MAY fill in this property in Constraint Sets basing on any additional information about Receivers. For example, if the user in any way informs NMOS Controller that Receiver D is more important than Receivers A, B, C and `urn:x-nmos:cap:meta:preference` of Receiver D states that Constraint Set 6 has higher preference than Constraints Sets 2, 3, 4, 5, then the consensus would be Constraint Sets 2, 3, 4, 5, 6 where Constraint Set 6 has the highest preference.

Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely while making the intersection.
```

docs/Behaviour - Client Side.md
N-Nagorny:
```suggestion
```

docs/Behaviour - Server Side.md
```diff
@@ -17,26 +17,26 @@ Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the
...
 
 ## State of Receiver
 
 A Receiver managed with IS-11 has the following states:
-- `No Transport File` when there's no active `transport_file`.
-- `OK` when this Receiver's active `transport_file` does not violate Receiver Capabilities.
-- `Receiver Capabilities Violation` when this Receiver's active `transport_file` violates Receiver Capabilities.
+- `no_transport_file` when there's no active `transport_file`.
+- `ok` when this Receiver's active `transport_file` does not violate Receiver Capabilities.
+- `receiver_capabilities_violation` when this Receiver's active `transport_file` violates Receiver Capabilities.
```
```suggestion
- `unknown` when this Receiver is inactive or it's active and it's impossible to say whether the Flow is compliant with this Receiver (e.g. it's activated without a `transport_file`).
- `compliant_flow` when this Receiver is active and the Flow is compliant with this Receiver (e.g. there's an active `transport_file` and it does not violate Receiver Capabilities).
- `non_compliant_flow` when this Receiver is active and the Flow is non compliant with this Receiver (e.g. there's an active `transport_file` which violates Receiver Capabilities).
```

examples/receiver-status-get-200.json
```diff
@@ -1,3 +1,3 @@
 {
-  "state": "OK"
+  "state": "ok"
```
N-Nagorny:
```suggestion
  "state": "compliant_flow"
```

docs/Behaviour - Server Side.md
```diff
@@ -17,26 +17,26 @@ Once a Sender accepts proposed Active Constraints, this Sender, the Flow and the
...
 
 ## Preventing restrictions violation
 
-At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST NOT transmit the Flow over the network. The Sender is REQUIRED to take the following actions:
+At any time if State of an active Sender becomes `active_constraints_violation`, the Sender MUST NOT transmit the Flow over the network. The Sender is REQUIRED to take the following actions:
 - MUST set the [IS-05][IS-05] `/active`'s `master_enable` property to `false`.
 - MUST update the [IS-04][IS-04] `subscription`'s `active` property to `false` (regardless of unicast or multicast).
 
-At any time if State of an active Receiver becomes `Receiver Capabilities Violation`, the Receiver MUST stop receiving the Flow by taking the following actions:
+At any time if State of an active Receiver becomes `receiver_capabilities_violation`, the Receiver MUST stop receiving the Flow by taking the following actions:
```
```suggestion
At any time if State of an active Receiver becomes `non_compliant_flow`, the Receiver MUST stop receiving the Flow by taking the following actions:
```

APIs/schemas/receiver-status.json
```diff
@@ -9,9 +9,9 @@
...
+        "receiver_capabilities_violation"
```
N-Nagorny:
```suggestion
        "unknown",
        "compliant_flow",
        "non_compliant_flow"
```

docs/Behaviour - Client Side.md
alabou: I think I would remove that paragraph ... because computing the intersection is much more complex than that. We could state that urn:x-nmos:cap:meta:label, urn:x-nmos:cap:meta:preference properties of constraint_set(s) MUST be ignored for the purpose of calculating the intersection and those constraint_set(s) with urn:x-nmos:cap:meta:enabled set to false MUST be ignored completely.

Ex. one Receiver indicate  height in range 720 <= height <= 2160 and another indicate height == 1080 ... The intersection is height 1080 but the constraint_set(s) are very different.

docs/Behaviour - Client Side.md
```diff
@@ -1,16 +1,16 @@
...
+If multiple Receivers have equal Constraint Sets, NMOS Controller MUST choose only one of them while making the intersection. If they differ only in `urn:x-nmos:cap:meta:preference`, then one with the highest priority MUST be taken.
 
 ## Dynamic format changes on Sender
 
 ### Changes which break Active Constraints
 
-NMOS Controller MUST track Sender's `subscription`'s `active` property during the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the request returns `Active Constraints Violation`, NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
+NMOS Controller MUST track Sender's `subscription`'s `active` property during the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
```
alabou, line 20: Update state name according to final naming convention ...

docs/Behaviour - Server Side.md
alabou: I would add "Capabilities" to all 3 cases "compliant with this Receiver Capabilities"

```suggestion
- `unknown` when this Receiver is inactive or it is active and it is impossible to say whether the Flow is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`).
- `compliant_flow` when this Receiver is active and the Flow is compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` and it does not violate the Receiver Capabilities).
- `non_compliant_flow` when this Receiver is active and the Flow is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities).
```

docs/Behaviour - Client Side.md
N-Nagorny: I'm about to remove it too. Can `urn:x-nmos:cap:meta:label` and `urn:x-nmos:cap:meta:preference` be used for setting `urn:x-nmos:cap:meta:preference` to Constraint Sets of the intersection? An example for `urn:x-nmos:cap:meta:preference` is given in a suggestion below.

docs/Behaviour - Server Side.md
```diff
@@ -12,24 +12,24 @@ Active Constraints are violated when at least one Parameter Constraint of each C
...
+- `ok` when this Receiver's active `transport_file` does not violate Receiver Capabilities.
+- `receiver_capabilities_violation` when this Receiver's active `transport_file` violates Receiver Capabilities.
 
 ## Preventing restrictions violation
 
-At any time if State of an active Sender becomes `Active Constraints Violation`, the Sender MUST become inactive. An inactive Sender in this state MUST NOT allow activations.
+At any time if State of an active Sender becomes `active_constraints_violation`, the Sender MUST become inactive. An inactive Sender in this state MUST NOT allow activations.
 
-At any time if State of an active Receiver becomes `Receiver Capabilities Violation`, the Receiver SHOULD become inactive. An inactive Receiver in this state SHOULD NOT allow activations.
+At any time if State of an active Receiver becomes `receiver_capabilities_violation`, the Receiver SHOULD become inactive. An inactive Receiver in this state SHOULD NOT allow activations.
```
```suggestion
At any time if State of an active Receiver becomes `non_compliant_flow`, the Receiver SHOULD become inactive. An inactive Receiver in this state SHOULD NOT allow activations.
```

docs/Behaviour - Server Side.md
alabou: There is a pending issue about the name "flow" versus "stream" ... If we go with "stream" we'll have to do a global replacement of "flow" for "stream" ...

APIs/schemas/receiver-status.json
N-Nagorny: Renamed to `compliant_flow`.

docs/Behaviour - Client Side.md
```diff
@@ -1,16 +1,23 @@
...
+
+`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates Constraint Set preference between Constraints Sets of the Receiver. NMOS Controller MUST remove values of this property before `PUT /constraints/active` and MAY use these values while making a decision about the intersection.
+
+`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Active Constraints indicates Constraint Set preference between Constraints Sets of the Active Constraints of the Sender. NMOS Controller MAY fill in this property in Constraint Sets basing on any additional information about Receivers. For example, if the user in any way informs NMOS Controller that Receiver D is more important than Receivers A, B, C and `urn:x-nmos:cap:meta:preference` of Receiver D states that Constraint Set 6 has higher preference than Constraints Sets 2, 3, 4, 5, then the consensus would be Constraint Sets 2, 3, 4, 5, 6 where Constraint Set 6 has the highest preference.
```
```suggestion
`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates Constraint Set preference between Constraint Sets of the Receiver. NMOS Controller SHOULD NOT propagate values of this property as is from Receiver Caps to Active Constraints and MAY use these values to make a decision about the intersection and determine preference of Constraint Sets of Active Constraints.

`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Active Constraints indicates Constraint Set preference between Constraints Sets of the Active Constraints of the Sender. NMOS Controller MAY fill in this property in Constraint Sets basing on any additional information about Receivers. For example, if the user in any way informs NMOS Controller that Receiver D is more important than Receivers A, B, C and `urn:x-nmos:cap:meta:preference` of Receiver D states that Constraint Set 6 has higher preference than Constraints Sets 2, 3, 4, 5, then the consensus would be Constraint Sets 2, 3, 4, 5, 6 where Constraint Set 6 has the highest preference.
```

docs/Behaviour - Server Side.md
```diff
@@ -12,24 +12,24 @@ Active Constraints are violated when at least one Parameter Constraint of each C
...
 
 ## State of Receiver
 
 A Receiver managed with IS-11 has the following states:
-- `No Transport File` when there's no active transport file.
-- `OK` when this Receiver's active transport file does not violate Receiver Capabilities.
-- `Receiver Capabilities Violation` when this Receiver's active transport file violates Receiver Capabilities.
+- `unknown` when this Receiver is inactive or it is active and it is impossible to say whether the Flow is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`).
+- `compliant_flow` when this Receiver is active and the Flow is compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` and it does not violate the Receiver Capabilities).
+- `non_compliant_flow` when this Receiver is active and the Flow is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities).
```
```suggestion
- `non_compliant_flow` when this Receiver is active and the Flow is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities). This state preserves until a state change even if the Receiver becomes inactive.
```

alabou: I would only state "If the Receiver becomes inactive this state is preserved until the next activation.". And to allow it the unknown state should indicate "`unknown` when this Receiver is inactive or active and it is impossible to say whether the Flow is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`)." The previous sentence implied that an inactive Receiver shall be at the unknown state.

```suggestion
- `unknown` when this Receiver is inactive or active and it is impossible to say whether the Flow is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`).
- `compliant_flow` when this Receiver is active and the Flow is compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` and it does not violate the Receiver Capabilities).
- `non_compliant_flow` when this Receiver is active and the Flow is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities). If the Receiver becomes inactive this state is preserved until the next activation.
```

docs/Behaviour - Client Side.md
garethsb:
```suggestion
`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates relative preference between Constraint Sets of that Receiver. The NMOS Controller MAY use these values to make a decision about the intersection and determine preference of Constraint Sets of Active Constraints. It SHOULD NOT propagate values of this property as is from Receiver Capabilities to Active Constraints.

`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Active Constraints indicates relative preference between Constraints Sets of the Active Constraints of the Sender. The NMOS Controller MAY fill in this property in Constraint Sets based on any additional information about Receivers. For example, if the user in any way informs the NMOS Controller that Receiver D is more important than Receivers A, B, C and `urn:x-nmos:cap:meta:preference` of Receiver D states that Constraint Set 6 has higher preference than Constraints Sets 2, 3, 4, 5, then the consensus would be Constraint Sets 2, 3, 4, 5, 6 where Constraint Set 6 has the highest preference.
```

I think the explanation of the example needs to be expanded as "more important" here seems to mean "sending the highest preference stream for Receiver D is so important that it doesn't matter if Receivers A, B, C would be not be capable of receiving this stream at all"?

PR #90, opened 2022-01-31
Add more detail on Input/Output description

N-Nagorny:


garethsb:
Rather than saying that it is unspecified, I would prefer there is additional explanation of the (unexpressed) relationship between IS-11 Inputs and IS-04 Sources, with examples of one Input being related to multiple Sources (and thus Senders, of course) and of multiple Inputs contributing to one Source and (potentially via the existing IS-04 `parents` mechanism) a Sender.

docs/Overview.md
```diff
@@ -35,11 +35,11 @@ and "OPTIONAL" in this documentation set are to be interpreted as described in [
 
 ### Input
 
-An Input is a unit consuming media data for providing it to Sender(s). Input may present metadata about media consuming capabilities to its upstream counterpart. Each Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
+An Input is a logical input connector which consumes video, audio and ancillary data for providing it to Sender(s). Relationship between the Input and such resources as a Source and a Flow is unspecified. The Input may present metadata about media consuming capabilities to its upstream counterpart. Each Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```
alabou: What about:
An Input is a logical input connector which consumes video, audio and ancillary data for providing it to Sender(s) through a number of Sources and Flows. The precise relationship between the Inputs and such Sources and a Flows is not specified as there may be IS-08 mapping element in the system. The Input may present metadata about media consuming capabilities to its upstream counterpart. Each Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].

N-Nagorny:
```suggestion
An Input is a logical input connector which consumes video, audio and ancillary data for providing it to Sender(s) through a number of Sources and Flows. The precise relationship between the Inputs and such Sources and Flows is not specified and there may be an IS-08 mapping element in the system. The Input may present metadata about media consuming capabilities to its upstream counterpart. Each Input is a resource and based on [Resource Core JSON Schema][Resource-Core-Schema].
```

N-Nagorny:
I tried to go through the path from one Input (video+audio) to two Senders (video and audio) and found https://specs.amwa.tv/is-04/releases/v1.3.1/APIs/schemas/with-refs/source_generic.html which states that a Source can originate multiple formats (video+audio).

Though https://specs.amwa.tv/is-04/releases/v1.3.1/docs/5.1._Data_Model_-_Identifier_Mapping.html#source-id says that Source ID "SHOULD change if <...> format URN changes between video, audio, data and mux variants" which looks a bit conflicting with the previous fact.

The vision mixer case looks pretty similar to the case of two Inputs contributing to a single Source. https://specs.amwa.tv/is-04/releases/v1.3.1/docs/5.1._Data_Model_-_Identifier_Mapping.html#sources

garethsb:
> found https://specs.amwa.tv/is-04/releases/v1.3.1/APIs/schemas/with-refs/source_generic.html which states that a Source can originate multiple formats (video+audio).

I don't think that says what you think. I believe it is just that the _source_generic.json_ schema is used to validate both video and mux sources, whereas data and audio sources both have a format-specific parameter (see _source_data.json_ and _source_audio.json_).

garethsb:
I think you **could** have one video+audio Input associated with one mux Source, and then a video Source and an audio Source both of which have the mux Source in their `parents`.

alabou:
> I think you **could** have one video+audio Input associated with one mux Source, and then a video Source and an audio Source both of which have the mux Source in their `parents`.

At this time Source and Flow can only have parents of their own type (a number of parent Sources for a Source, a number of parent Flows for a Flow).  Currently there is no method for associating a Source to an IS-11 Input. I think this is not in the IS-11 mandate right now. A possible future solution would be to have a tags "parent_input" on a Source that identifies the IS-11 Input ID. The "mux" source of the quoted example could refer to a parent IS-11 Input that way. But I think that the "mux" source is not required and that both the audio and video Source could have the same IS-11 Input as an parent_input. Multiple Sources could refer to the same IS-11 Input as a parent_input.

docs/Data Model.md
```diff
@@ -10,6 +10,8 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 SHOULD be associated with 1 or more Input resources and Receivers managed with IS-11 SHOULD be associated with 1 or more Output resources.
 
+Inputs/Outputs carrying multiple media formats at a time (e.g. video and audio) are RECOMMENDED to be associated with multiple Senders/Receivers of given formats.
+
```
N-Nagorny:
```suggestion
Inputs/Outputs carrying multiple media formats at a time (e.g. video and audio) MAY be associated with multiple Senders/Receivers of given formats.

```

N-Nagorny:
2022/02/21 meeting: we're closing it with undefined relationship between Inputs and Senders but it's a task for the future to decide whether we need to tackle it in IS-11.

PR #92, opened 2022-02-09
Fix Identifier Persistence

### LINKED ISSUE #91: About Input/Output ID
The current IS-11 language about Input ID and Output ID describes a Sender as a parent of an Input and a Receiver as a parent of an Output ... I think think this is wrong and possibly a legacy of a previous model of IS-11.

An Input ID should not change if the ID of a Sender associated to the Input changes ...
An Output ID should not change if the ID of a Receiver associated to the Output changes ... 

The IS-04 model is based on this "Owned by: Device ID SHOULD change if: Device ID changes ... " Input and Output are owned by Device ID ...

I would suggest to revise this section of the IS-11 specification.

N-Nagorny:
Resolves #91.

### DIFF
--- a/docs/Data Model.md
+++ b/docs/Data Model.md
@@ -36,7 +36,7 @@ Owned by:
 
 SHOULD change if:
 
-- Sender ID changes
+- Device ID changes
 - A different physical interface (such as an SDI input) is mapped to originate the essence
 
 SHOULD NOT change if:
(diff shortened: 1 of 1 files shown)


docs/Data Model.md
```diff
@@ -37,16 +37,16 @@ Owned by:
 SHOULD change if:
 
 - Sender ID changes
```
alabou: change "Sender" for "Device"

```suggestion
- Device ID changes
```

PR #96, opened 2022-04-03
Replace "contains" and "minContains" with draft-04 means

N-Nagorny:


N-Nagorny:
As far as the JSON Schemas in the matter are designed for humans at first, we can't sacrifice readability, so the group decided to express the required supported parameter constraints with spec language means. Closed.

N-Nagorny:
As far as the JSON Schemas in the matter are designed for humans at first, we can't sacrifice readability, so the group decided to express the required supported parameter constraints with spec language means. Closed.

PR #99, opened 2022-04-15
Sender/Receiver without Inputs/Outputs

N-Nagorny:


docs/Behaviour - Client Side.md
```diff
@@ -21,6 +21,6 @@ NMOS Controller MUST track Sender's `subscription`'s `active` property during an
 
 ### Changes which do not break Active Constraints
 
-Sender's changes which follow the Active Constraints do not effect the `active` property so NMOS Controller MUST track version of Inputs, the Source, the Flow and the Sender. If any of them has changed, NMOS Controller SHOULD `PATCH /staged` of all the Receivers with new `transport_file` (if it's used for the IS-05 connection) and immediate activation.
+Sender's changes which follow the Active Constraints do not effect the `active` property so NMOS Controller MUST track version of Inputs, the Source, the Flow and the Sender. If transport file of the Sender has changed, NMOS Controller SHOULD `PATCH /staged` of all the Receivers with new `transport_file` (if it's used for the IS-05 connection) and immediate activation.
```
```suggestion
Sender's changes which follow the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` (if it's used for the IS-05 connection) and immediate activation.
```

APIs/schemas/sender-status.json
```diff
@@ -12,8 +12,8 @@
         "unconstrained",
         "constrained",
         "active_constraints_violation",
-        "no_signal",
-        "awaiting_signal"
+        "no_stream",
+        "awaiting_stream"
```
N-Nagorny:
```suggestion
        "no_signal",
        "awaiting_signal"
```

docs/Behaviour - Server Side.md
```diff
@@ -32,10 +32,10 @@ A Sender managed with IS-11 has the following states:
 - `unconstrained` when Active Constraints of this Sender is an empty `constraint_sets` array.
 - `constrained` when the Active Constraints are satisfied.
 - `active_constraints_violation` when Active Constraints are violated.
-- `no_signal` when there is no signal from Inputs associated with this Sender.
-- `awaiting_signal` when the signal from Inputs associated to this Sender is transitioning and the signal is not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
+- `no_stream` when there is no stream provided to this Sender.
+- `awaiting_stream` when the stream provided to this Sender is transitioning and not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
```
```suggestion
- `no_signal` when there is no signal provided to this Sender.
- `awaiting_signal` when the signal provided to this Sender is transitioning and not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
```

docs/Behaviour - Server Side.md
```diff
@@ -32,10 +32,10 @@ A Sender managed with IS-11 has the following states:
...
 
-`no_signal` and `awaiting_signal` states MUST depend only on the signal from the Inputs. One of the other states MUST be used when the signal is present, stable and can be validated against Active Constraints.
+`no_stream` and `awaiting_stream` states MUST depend only on the stream provided to the Sender (e.g. from Inputs associated with this Sender if used). One of the other states MUST be used when the stream is present, stable and can be validated against Active Constraints.
```
```suggestion
`no_signal` and `awaiting_signal` states MUST depend only on the signal provided to the Sender (e.g. from Inputs associated with this Sender if used). One of the other states MUST be used when the signal is present, stable and can be validated against Active Constraints.
```

docs/Data Model.md
```diff
@@ -8,11 +8,11 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 ![Extended IS-04 Model Diagram](../docs/images/is-11-data-model.png)
 
-The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 SHOULD be associated with 1 or more Input resources and Receivers managed with IS-11 SHOULD be associated with 1 or more Output resources.
+The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 are associated with 0 or more Input resources and Receivers managed with IS-11 are associated with 0 or more Output resources.
```
N-Nagorny:
```suggestion
The IS-04 hierarchy is preserved. Additionally, there are Input and Output resources introduced. Senders managed with IS-11 are associated with 0 or more Input resources and Receivers managed with IS-11 are associated with 0 or more Output resources.
```

docs/Behaviour - Client Side.md
```diff
@@ -21,6 +21,6 @@ NMOS Controller MUST track Sender's `subscription`'s `active` property during an
 
 ### Changes which do not break Active Constraints
 
-Sender's changes which follow the Active Constraints do not effect the `active` property so NMOS Controller MUST track version of Inputs, the Source, the Flow and the Sender. If any of them has changed, NMOS Controller SHOULD `PATCH /staged` of all the Receivers with new `transport_file` (if it's used for the IS-05 connection) and immediate activation.
+Sender's changes which follow the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` (if it's used for the IS-05 connection) and immediate activation.
```
garethsb:
```suggestion
Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
```

alabou (approved):

docs/Data Model.md
```diff
@@ -62,7 +62,7 @@ Owned by:
 SHOULD change if:
 
 - Device ID changes
-- A different physical interface (such as an SDI output) is mapped to consume the essence
+- A different physical interface (such as an SDI output) is mapped to consume the stream
```
alabou: Why "stream" here and "signal" above?

docs/Data Model.md
```diff
@@ -8,11 +8,11 @@ In order to ensure consistent behaviour in a facility employing this identity, s
 
 ![Extended IS-04 Model Diagram](../docs/images/is-11-data-model.png)
 
-The IS-04 hierarchy is preserved. Additionally, Senders managed with IS-11 SHOULD be associated with 1 or more Input resources and Receivers managed with IS-11 SHOULD be associated with 1 or more Output resources.
+The IS-04 hierarchy is preserved. Additionally, there are Input and Output resources introduced. Senders managed with IS-11 are associated with 0 or more Input resources and Receivers managed with IS-11 are associated with 0 or more Output resources.
```
alabou, line 11: "are" of "MAY be" associated ... ?

garethsb: MAY expresses an OPTIONAL item. If it said 1 or more, I think that MAY would clearly be inappropriate since making this optional would mean some implementations could have 0 and that would be permitted. So here, where we are saying 0 or more, I think it is probably better not to use conformance language at all, and use "are" or "can be"?

docs/Data Model.md
garethsb: Ah good point, it's not the stream that was received at the Receiver, it's the signal that the Output is, hmm, consuming seems wrong, as does originating on the Input side?

N-Nagorny: My bad, here should be "signal".
```suggestion
- A different physical interface (such as an SDI output) is mapped to consume the signal
```
It is a different signal, but should we emphasize this? If we allow Outputs not associated with Receivers, I don't see how we can define what is the signal that the Output consumes. It may be any signal from anywhere, isn't it?

PR #103, opened 2022-05-23
Add distinction between signal and Essence

N-Nagorny:


APIs/schemas/input.json
```diff
@@ -21,7 +22,15 @@
...
           "type": "boolean"
         },
-        "effective_edid": { "$ref": "edid.json" }
+        "effective_edid": { "$ref": "edid.json" },
+        "signal_state": {
+          "type": "string",
+          "enum": [
+            "no_signal",
+            "awaiting_signal",
+            "signal_is_present"
```
garethsb: Or `signal_stable` ?

```suggestion
            "signal_present"
```

docs/Behaviour - Server Side.md
```diff
@@ -32,10 +32,10 @@ A Sender managed with IS-11 has the following states:
 - `unconstrained` when Active Constraints of this Sender is an empty `constraint_sets` array.
 - `constrained` when the Active Constraints are satisfied.
 - `active_constraints_violation` when Active Constraints are violated.
-- `no_signal` when there is no signal provided to this Sender.
-- `awaiting_signal` when the signal provided to this Sender is transitioning and not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
+- `no_essence` when the associated Source is not providing any Essence for this Sender to transmit, which, if the Sender has associated Inputs, can mean the one or more of these Inputs have no incoming signal.
```
garethsb:
```suggestion
- `no_essence` when the associated Source is not providing any Essence for this Sender to transmit, which, if the Sender has associated Inputs, can mean that one or more of these Inputs have no incoming signal.
```

docs/Behaviour - Server Side.md
```diff
@@ -32,10 +32,10 @@ A Sender managed with IS-11 has the following states:
...
+- `awaiting_essence` when the Essence, which the associated Source is providing to this Sender, is transitioning and not stable yet. This is a transitional state until one of the previous ones can be established. It is RECOMMENDED to not update resources such as Flow and Source associated with a Sender while in this state.
 
-`no_signal` and `awaiting_signal` states MUST depend only on the signal provided to the Sender (e.g. from Inputs associated with this Sender if used). One of the other states MUST be used when the signal is present, stable and can be validated against Active Constraints.
+`no_essence` and `awaiting_essence` states MUST depend on the state of Input(s) associated with this Sender when the Essence is based exclusively on the signal from these Input(s).
```
garethsb: Should it say "`signal_state` of the Inputs" maybe?

```suggestion
`no_essence` and `awaiting_essence` states MUST depend on the state of the Inputs associated with this Sender when the Essence is based exclusively on the signal from these Inputs.
```

examples/input-properties-get-200.json
```diff
@@ -5,5 +5,6 @@
   "description": "Video card in the PC placed in the office",
   "tags": {},
   "connected": true,
-  "edid_support": true
+  "edid_support": true,
+  "signal_state": "signal_is_present"
```
garethsb:
```suggestion
  "signal_state": "signal_present"
```

garethsb (commented):
Now that we have `signal_state` of Inputs, do we need `awaiting_essence` state of Senders? Would it make sense to just use `no_essence` in this case? Maybe it's still important, I'm not sure.

APIs/schemas/input.json
alabou: IS-11 must work without Inputs. This state relates to the transitional period when an upstream device changes its signal (resolution, grain rate, etc). During the reacquisition of the signal a device would indicate awaiting_essence.  The state no_essence indicates that there us no signal at all …

examples/input-properties-get-200.json
alabou: Should there be a state “signal_not_present”  or have it a boolean?

N-Nagorny: The `enum` in the schema lists `no_signal`, `awaiting_signal` and `signal_is_present` as it was for a Sender.

APIs/schemas/input.json
N-Nagorny: Separating the state of an Input from the state of a Sender doesn't prevent us from having IS-11 without Inputs. `awaiting_signal` for an Input makes sense for the case you've described. The question is whether `awaiting_essence` makes sense for a Sender. As far as there may be a Sender completely depending from its Inputs, I think it makes sense. From the client's perspective PUTting Active Constraints causes the reacquisition of the signal and therefore the reacquisition of the Essence in this case.

PR #105, opened 2022-05-30
Add "debug" property to Statuses

N-Nagorny:


N-Nagorny:
Status of Output may be useful as well. It should be added. `debug` property is needed, states are `no_signal` and `signal_present`.

docs/Behaviour - Server Side.md
```diff
@@ -26,7 +26,24 @@ A Sender MUST present the list of supported parameter constraints in `/supported
...
+An Input has the following states:
+- `no_signal` when there is no incoming signal to this Input.
+- `awaiting_signal` when the incoming signal is not stable. This is a transitional state until one of the other ones can be established.
+- `signal_present` when the incoming signal is present and stable.
+
+Debug information expressed through `debug` property SHOULD assist `no_signal` state and MAY assist the other states.
+
+## Status of Output
+
+An Output has the following states that describe the signal provided by associated Receiver(s) to this Output:
```
garethsb: Should this be a little more like the explanation of `no_essence` at the Sender?

Something like:

An Output has the following states that describe the signal at this Output:
- `no_signal` when there is no signal, which, if the Output has associated Receivers, can mean that one or more of these Receivers do not have a compliant active stream.

N-Nagorny: Thank you! I added it.

docs/Behaviour - Server Side.md
```diff
@@ -26,24 +26,45 @@ A Sender MUST present the list of supported parameter constraints in `/supported
...
 A Receiver managed with IS-11 has the following states:
 - `unknown` when it is not possible to say whether the active stream is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`).
 - `compliant_stream` when the active stream is compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` and it does not violate the Receiver Capabilities).
 - `non_compliant_stream` when the active stream is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities). If the Receiver becomes inactive this state is preserved until the next activation.
 
+Debug information expressed through `debug` property SHOULD assist `unknown` and `non_compliant_stream` states and MAY assist the other states.
+
+## Status of Output
+
+An Output has the following states that describe the signal provided by associated Receiver(s) to this Output:
```
garethsb: The relationship to Receivers is covered by following bullet point.

```suggestion
An Output has the following states that describe the signal at this Output:
```

docs/Behaviour - Server Side.md
```diff
@@ -26,24 +26,45 @@ A Sender MUST present the list of supported parameter constraints in `/supported
...
 - `unknown` when it is not possible to say whether the active stream is compliant with Receiver Capabilities of this Receiver (e.g. it is activated without a `transport_file`).
 - `compliant_stream` when the active stream is compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` and it does not violate the Receiver Capabilities).
 - `non_compliant_stream` when the active stream is non compliant with Receiver Capabilities of this Receiver (e.g. there is an active `transport_file` which violates the Receiver Capabilities). If the Receiver becomes inactive this state is preserved until the next activation.
 
+Debug information expressed through `debug` property SHOULD assist `unknown` and `non_compliant_stream` states and MAY assist the other states.
+
+## Status of Output
+
+An Output has the following states that describe the signal at this Output:
+- `no_signal` when there is no signal, which, if the Output has associated Receivers, can mean that one or more of these Receivers do not have a compliant incoming stream.
```
```suggestion
- `no_signal` when this Output is not producing any signal to a downstream counterpart (e.g. if the Output has associated Receivers, it may mean that one or more of these Receivers do not have a compliant incoming stream).
```

PR #109, opened 2022-11-08
Fix the language about the NMOS Controller makeing IS-05 connections (active/inactive Sender, keep/change configuration of Sender, use real/fictious Receivers)

alabou:
The actual language about the NMOS Controller is too restrictive and does not cover the cases where the Sender is already active and Receivers are to be connected to it without changing the configuration of the Sender. There are also scenarios involving not real actual Receivers of the NMOS environment but targeted fictitious Receivers that are use to preset the Sender.

The new language attempt to capture the active/inactive state of the Sender selected by the NMOS Controller and the fact that the configuration of such Sender is to be kept or changed. Finally to allow an NMOS Controller to configure a Sender without considering real effective Receivers the new language allow the use of target fictitious Receivers.

It seems reasonable to allow more flexibility in the use of IS-11 by an NMOS Controller.


garethsb:
I like the intention.

Although I understand what you mean, I'm not sure whether the terminology of "effective/real" and "target/fictitious" is right yet. Maybe we need a naming brainstorm on Slack. 😄 

The duplication between the "If changing the configuration" and the "If keeping the configuration" paragraphs makes each one easy to read, but I wonder if some of the commonality can be factored out.

alabou:
Thanks for your comment. I agree that the "effective/real" and "target/fictitious" expressions are not the best .. Initially I just wanted to express the objectives of the modification and I would appreciate your comments about making the language better. Initially we must agree about the objectives and then fix the language.

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,13 @@
...
+Before making an [IS-05][IS-05] connection, an NMOS Controller chooses a Sender and a number of Receivers. The NMOS Controller MAY use effective Receivers (real devices) or target Receivers (fictitious devices). The NMOS Controller MAY choose to connect the Receivers to an active Sender or an inactive Sender. If the Sender is active, the NMOS Controller MAY keep the actual configuration of the Sender or deactivate the Sender and change its configuration. If the Sender is inactive, the NMOS Controller MAY keep the actual configuration of the Sender or change its configuration.
```
SoloAs, line 5:
```suggestion
Before making an [IS-05][IS-05] connection, an NMOS Controller chooses a Sender and a number of Receivers. The NMOS Controller MAY choose to connect the Receivers to an active Sender or an inactive Sender. If the Sender is active, the NMOS Controller MAY keep the actual configuration of the Sender or deactivate the Sender and change its configuration. If the Sender is inactive, the NMOS Controller MAY keep the actual configuration of the Sender or change its configuration.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,13 @@
...
+
+If keeping the configuration of the Sender, the NMOS Controller MUST `GET /constraints/supported` and `GET /constraints/active` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such processing of them that they can be utilized for building Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, the NMOS Controller SHOULD inform the user about it. If the resulting `constraint_sets` is not a subset of the active constraints of the sender the NMOS Controller SHOULD inform the user about it. After that the NMOS Controller SHOULD make the IS-05 connection of the Receivers.
+
+After breaking these connections via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender after making it inactive.
```
SoloAs, line 11:
```suggestion
After breaking these connections via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender after making it inactive.

In order to change the configuration of the Sender without establishing an IS-05 connection, NMOS Controller MUST use `PUT /constraints/active` endpoint. If the Sender is active, NMOS Controller SHOULD deactivate it first.
```

alabou: I like that way of introducing the alternate approach "In order to change the configuration of the Sender without establishing an IS-05 connection".

With this phrasing and the "MAY choose to connect the Receivers to an active Sender or an inactive Sender" along with the accompanying text seems to cover all the possibilities.

It would be nice to look at the final text as a single piece to make sure everything is there.

docs/Behaviour - Client Side.md
N-Nagorny: The spec doesn't prohibit changing the configuration of an active Sender, so we can remove this distinction or emphasize that deactivating such Sender MAY be required and depends on the Sender itself.

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,13 @@
...
+
+If keeping the configuration of the Sender, the NMOS Controller MUST `GET /constraints/supported` and `GET /constraints/active` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such processing of them that they can be utilized for building Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, the NMOS Controller SHOULD inform the user about it. If the resulting `constraint_sets` is not a subset of the active constraints of the sender the NMOS Controller SHOULD inform the user about it. After that the NMOS Controller SHOULD make the IS-05 connection of the Receivers.
```
N-Nagorny, line 9: >is not a subset

Maybe requires a definition of subset. Here is one that Gareth and me formulated for the nmos-cpp PR:
>Constraint Set B is a subset of Constraint Set A if all Parameter Constraints of Constraint Set A are present in Constraint Set B, and for each Parameter Constraint that is present in both, the Parameter Constraint of Constraint Set B is a subconstraint of the Parameter Constraint of Constraint Set A.

>Constraint B is a subconstraint of Constraint A if:
>1. Constraint B has enum keyword when Constraint A has it and enumerated values of Constraint B are a subset of enumerated values of Constraint A
>2. Constraint B has enum or minimum keyword when Constraint A has minimum keyword and allowed values of Constraint B are greater than minimum value of Constraint A
>3. Constraint B has enum or maximum keyword when Constraint A has maximum keyword and allowed values of Constraint B are less than maximum value of Constraint A

docs/Behaviour - Client Side.md
garethsb: Agree with Nikita. Also I prefer "current configuration" to "actual configuration".

docs/Behaviour - Client Side.md
garethsb: Is that recommendation to deactivate before changing the active constraints covered anywhere else in the spec so far?

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,13 @@
...
+
+If changing the configuration of the Sender, the NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such processing of them that they can be utilized for building Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, the NMOS Controller SHOULD inform the user about it. After that the NMOS Controller MUST `PUT /constraints/active` to the Sender and SHOULD make the IS-05 connections of the Receivers if the Constraints have been applied succesfully to the Sender.
```
garethsb, line 7: > If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, the NMOS Controller SHOULD inform the user about it.

Does empty `constraint_sets` mean no Constraint Sets at all, i.e. it was impossible to come up with any Constraint Set that would satisfy all of the Receivers? If so, is SHOULD enough? How would the Controller be expected to proceed if it cannot make the connection to all of the Receivers that the User requested? I think the answer is that something like this:

> If the processing of the Receiver Capabilities results in an empty `constraint_sets`, the NMOS Controller SHOULD inform the User that not all of the specified Receivers can be satisfied at the same time.

garethsb, line 7: "User" is a defined NMOS term now - https://specs.amwa.tv/nmos/branches/main/docs/Glossary.html#user - so we could capitalize it like other terms if we like?

docs/Behaviour - Client Side.md
alabou: I made a mistake in the sentence .. It shall be the reverse "If the active constraints are not a subset of the resulting `constraint_sets`" ... that is: the new constraint_sets must not be more constrained than the actual ones because it would mean that a stream/flow produces by the active sender would not be compliant with the new receiver(s).

My definition of a subset is different ... Constraint Sets A is a subset of Constraint Sets B if a stream/flow compliant with CS-A is also compliant with CS-B. I don't think we should define the complete algorithm that ensure that CS-A is a subset with CS-B but dictate what such subset represent.

garethsb: Hm, yeah, use of _subset_ at this point is interesting, given we've been over the spec language before...

The current statement above that a Controller "make **_such processing of them that they can be utilized for building_** Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender" is quite vague but could be implemented as _intersection_ of the Receivers' `constraint_sets`. Effectively, Nikita and I had to come up with the above subset/intersection definition when implementing this.

When we have access to the Sender's own capabilities... which we didn't quite finish off specifying (https://github.com/AMWA-TV/is-11/issues/63).... the Controller would sensibly consider including those in the processing/subset to ensure it will get a `200` not a `422`!

garethsb: Hm, actually, isn't this whole paragraph a bit confused... The fact that the "processing of the Receiver Capabilities" results in a subset or not doesn't mean that the Sender's current configuration will match all those Receivers, since the current stream might be within the Sender's current active constraints but not in the subset supported by the new Receivers?
Unless the Controller _applies_ these new active constraints with a `PUT` request, it cannot guarantee either way?

Unless I've misunderstood, in order to guarantee to satisfy the new Receivers, the Controller MUST apply the new active constraints unless the _current_ ones are a subset of the _new_ ones?

garethsb: Aha, yes, I see @alabou made same comment... I agree with your idea!

garethsb: > If the resulting `constraint_sets` is not a subset of the active constraints of the sender the NMOS Controller SHOULD inform the user about it.

So, putting this the right way round, avoiding "subset", clarifying the rationale for informing the user, and expanding, something like:

> If there can be streams compliant with the current active constraints of the Sender that would not be compliant with the resulting `constraint_sets`, the NMOS Controller SHOULD inform the User that the Receivers cannot be accommodated without replacing the active constraints.

Long winded, so maybe needs breaking into several sentences...

docs/Behaviour - Client Side.md
alabou: Is this read better?
> If the processing of the Receiver Capabilities results in an empty constraint_sets, the NMOS Controller SHOULD inform the User that all of the specified Receivers cannot be satisfied at the same time.

I think we should keep the SHOULD to allow the controller to decide in some scenarios ...

docs/Behaviour - Client Side.md
alabou: OK with that wording.

N-Nagorny:
Closed in favor of #115

PR #110, opened 2023-02-06
Added device_id to input/output properties, updates schemas and examp…

SoloAs:
Input/Output information about its device_id. Parent issue: https://github.com/AMWA-TV/is-11/issues/108

### DIFF
--- a/APIs/schemas/input.json
+++ b/APIs/schemas/input.json
@@ -10,7 +10,8 @@
       "required": [
         "connected",
         "edid_support",
-        "status"
+        "status",
+        "device_id"
       ],
       "properties": {
         "base_edid": { "$ref": "edid.json" },
@@ -42,6 +43,11 @@
               "type": "string"
             }
           }
+        },
+        "device_id": {
+          "description": "Device ID which this Input forms part of. This attribute is used to ensure referential integrity by registry implementations.",
+          "type": "string",
+          "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
         }
       }
     }
--- a/APIs/schemas/input.json
+++ b/APIs/schemas/input.json
@@ -45,7 +45,7 @@
           }
         },
         "device_id": {
-          "description": "Device ID which this Input forms part of. This attribute is used to ensure referential integrity by registry implementations.",
+          "description": "Device ID which this Input forms part of",
           "type": "string",
           "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
         }
--- a/APIs/schemas/output.json
+++ b/APIs/schemas/output.json
@@ -10,7 +10,8 @@
       "required": [
         "connected",
         "edid_support",
-        "status"
+        "status",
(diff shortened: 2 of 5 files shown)


docs/Interoperability.md
```diff
@@ -57,7 +57,7 @@ The UUIDs used to advertise Senders and Receivers in the Stream Compatibility Ma
 
 In order to prevent unnecessary polling of the Stream Compatibility Management API, changes are signaled via the IS-04 versioning mechanism.
 
-When properties of any Input/Output are changed, then the `version` attribute of the relevant IS-04 Device MUST be incremented.
+When properties of any Input/Output are changed, then the `version` attribute of the relevant IS-04 Device MUST be incremented. For this purpose Inputs/Outputs are linked to the corresponding devices via "device_id" property.
```
garethsb: Firstly, small suggested change:

```suggestion
When properties of any Input/Output are changed, then the `version` attribute of the relevant IS-04 Device MUST be incremented.
Inputs/Outputs identify the corresponding Devices via the `device_id` property.
```

APIs/schemas/input.json
```diff
@@ -42,6 +43,11 @@
               "type": "string"
             }
           }
+        },
+        "device_id": {
+          "description": "Device ID which this Input forms part of. This attribute is used to ensure referential integrity by registry implementations.",
```
garethsb: I don't think this is related to registry implementations, so unless we can think of better explanation, I suggest deleting second sentence.

```suggestion
          "description": "Device ID which this Input forms part of",
```

APIs/schemas/output.json
```diff
@@ -40,7 +41,12 @@
               "type": "string"
             }
           }
-        }
+        },
+        "device_id": {
+          "description": "Device ID which this Output forms part of. This attribute is used to ensure referential integrity by registry implementations.",
```
garethsb: I don't think this is related to registry implementations, so unless we can think of better explanation, I suggest deleting second sentence.

```suggestion
          "description": "Device ID which this Output forms part of",
```

docs/Interoperability.md
garethsb: Secondly, not related to the PR, but a question about the previous sentence "when properties of any Input/Output are changed, ..."

Does "properties" mean formally the `/properties` endpoint, so that whenever e.g. `status` changes, a Device `version` update is required?

APIs/schemas/output.json
alabou: Inputs/Outputs are not stored in the  registry. So I agree to remove this sentence "This attribute is used to ensure referential integrity by registry implementations."

docs/Interoperability.md
alabou: My understanding is that "properties" refer to the general properties of an Input/Output object referenced by a given id, not only what is being accessible through /properties. This requirement relates to the fact that inputs and outputs are not in the registry such that a Controller would know to check the device input/output state when polling the device version ...

SoloAs: Hi everyone. Sorry for the late response, wasn't working last couple days due to sickness.

Thank you for the proposal, word "identify" fits here better than "is linked", committing this one.
As for the not PR-related question, my understanding pretty much reflects one of Alains: properties are there to track changes of input and output objects without adding them to the registry directly.

APIs/schemas/output.json
SoloAs: Agreed

PR #111, opened 2023-02-15
Add "base_edid_changeable" to Input

N-Nagorny:
Add "base_edid_changeable" to Input

### DIFF
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -275,7 +275,7 @@ mediaType: application/json
                 application/octet-stream:
                   type: file
             204:
-              description: There is no Base EDID set or the Input doesn't support Base EDID at all.
+              description: There is no Base EDID set or the Input doesn't support Base EDID or EDID at all.
             307:
               description: A temporary redirect to the Base EDID file.
             404:
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -275,7 +275,7 @@ mediaType: application/json
                 application/octet-stream:
                   type: file
             204:
-              description: There is no Base EDID set.
+              description: There is no Base EDID set or the Input doesn't support Base EDID at all.
             307:
               description: A temporary redirect to the Base EDID file.
             404:
@@ -301,7 +301,7 @@ mediaType: application/json
               body:
                 type: !include schemas/error.json
             405:
-              description: Returned when the Input doesn't support changing Base EDID or doesn't support EDID at all.
+              description: Returned when the Input doesn't support Base EDID or doesn't support EDID at all.
               body:
                 type: !include schemas/error.json
             423:
@@ -324,7 +324,7 @@ mediaType: application/json
               body:
                 type: !include schemas/error.json
             405:
-              description: Returned when the Input doesn't support changing Base EDID or doesn't support EDID at all.
+              description: Returned when the Input doesn't support Base EDID or doesn't support EDID at all.
               body:
                 type: !include schemas/error.json
             423:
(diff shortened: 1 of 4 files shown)


N-Nagorny:
I made this property required, so in case of Input without EDID support it's always `false`.

garethsb:
I wonder whether this should be explicitly documented as implying that `PUT` or `DELETE` `/inputs/{inputId}/edid/base` return `405` Method Not Allowed?

N-Nagorny:
@garethsb, thank you, it's a good point.

I also removed
>There is no Base EDID at the initial state.

because I see nothing in the spec text that prevents an Input to have a default read-only Base EDID.

alabou:
I disagree with removing that sentence "There is no Base EDID at the initial state.". You are changing the specification and IS-11 will never be completed if we keep on changing the specification.

You cannot remove that sentence. If there is no base EDID it is the manufacturer EDID that is presented to the source device. "If Base EDID is not set, Effective EDID is built on the basis of a default Base EDID defined for the Input by the manufacturer."

alabou:
I would not use the name base_edid_changeable but base_edid_support to better match edid_support that we already have. This is just a name so I'm ok with any name as along as it does not imply changing the actual behavior implied by the specification.

N-Nagorny:
@alabou, hm, let me think about it. In case of an Input that doesn't allow changing its Base EDID, it still uses some Base EDID under the hood. This Base EDID can be fetched via `/edid/effective` when Active Constraints are empty.

Does it make sense to return this Base EDID from `/edid/base` of such Input? Maybe, no. We can keep the removed sentence, rename the property to `base_edid_support` and prescribe `GET /edid/base` to return 204 if we consider Base EDID as a resource controlled by the user and not changed internally.

But in this case, I would replace "a default Base EDID" in "If Base EDID is not set, Effective EDID is built on the basis of a default Base EDID defined for the Input by the manufacturer." with something to distinct the manufacturer EDID and Base EDID.

alabou:
I would be ok with a modification to "If Base EDID is not set, Effective EDID is built on the basis of a default Base EDID defined for the Input by the manufacturer." for "If Base EDID is not set, Effective EDID is built on the basis of a default EDID defined for the Input by the manufacturer."

"GET /edid/base to return 204" it is already the case in the RAML if there is no Base EDID set.

N-Nagorny:
Added a commit removing the citing of RAML but preserving the hint itself.

APIs/StreamCompatibilityManagementAPI.raml
```diff
@@ -275,7 +275,7 @@ mediaType: application/json
                 application/octet-stream:
                   type: file
             204:
-              description: There is no Base EDID set.
+              description: There is no Base EDID set or the Input doesn't support Base EDID or EDID at all.
```
garethsb, line 276: Does the use of `204` No Content and `405` Method Not Allowed when the Input doesn't support Base EDID or EDID at all seem a little inconsistent between methods of `/edid`, `/edid/base` and `/edid/effective`?

If EDID and Base EDID _are_ supported, then `204` No Content at `/edid/base` to indicate no Base EDID is set seems appropriate.

But the use of `204` to indicate EDID or Base EDID is not supported seems a bit odd.

The response to `/edid` normally indicates the sub-endpoints... if it returns `204` No Content, should they respond successfully? Also seems a bit strange to respond with a success code to `GET` `/edid/base` and `/edid/effective`, when `PUT` and `DELETE` are specified to return `405` Method Not Allowed to indicate the same thing...

garethsb: Even if it is preferable to keep `204` to indicate both not supported and supported-but-not-present on the `GET` requests for `/edid/base` and `/edid/effective`, would it still be a good idea to remove the `204` response from the parent `/edid` endpoint and always return `200` with the array of sub-endpoints?

alabou: I agree to only support 200 to return the API hierarchy of end-points and possibly an extra 404 when there is an {id} in the path. (This is what we agreed in the meeting on Feb 20, 2023)

garethsb (approved):
LGTM

PR #113, opened 2023-02-27
Add OPTIONS and other common NMOS API requirements

### LINKED ISSUE #112: Missing OPTIONS for CORS preflight
The endpoints that support modifying methods need to support CORS preflight, so the OPTIONS method needs to explicitly included in the RAML for:

* `/senders/{senderId}/constraints/active`
* `/inputs/{inputId}/edid/base`

Plenty of examples of this in other specs. Sorry I missed this before now...

N-Nagorny:
Resolves #112.

### DIFF
--- a/docs/APIs.md
+++ b/docs/APIs.md
@@ -10,9 +10,9 @@ _(c) AMWA 2023, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
 
 The Stream Compatibility Management API is specified using:
 
-* The following sub-sections describing common NMOS API properties.
-* [RAML](http://raml.org/) documents and [JSON schemas](http://tools.ietf.org/html/draft-zyp-json-schema-04) in the [APIs](../APIs/) folder.
-* Normative text in files in the [docs](../docs) directory.
+- The following sub-sections describing common NMOS API properties.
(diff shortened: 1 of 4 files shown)


docs/APIs.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+The `<api selector>` identifier MUST conform to the following regex pattern:
+
+```
+^[a-zA-Z0-9\-_]+$
+```
+
+This permits the use of a UUID, such as an associated IS-04 Device `id`, in the usual format or succinct local identifiers such as `slot2B`.
+
```
garethsb: Do we need this feature of IS-08? Not convinced we do.

garethsb (commented):
Addition of `OPTIONS` to the RAML looks good to me.

One question about need for API selectors.

Regarding rest of APIs and Server Side Implementation, I would like to review the other near copies of this text in e.g. IS-05 as well as IS-08 where you sourced it, to check we are using the latest revision, since we don't have an NMOS "core" spec to point to.

docs/APIs.md
garethsb: Wow, that (8f74f56) was fast, @N-Nagorny! LGTM.

garethsb (approved):
LGTM

PR #114, opened 2023-03-06
Is 05 controller language

SoloAs:
- PR updates with notes from other participants


docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
+Before making an [IS-05][IS-05] connection, an NMOS Controller chooses a Sender and a number of Receivers. The NMOS Controller MAY keep the current configuration of the Sender or change its configuration. In case of changing configuration of an active Sender, deactivating such Sender MAY be required and depends on the Sender itself.
```
garethsb, line 5: Is the last sentence referring to the possibilities covered in the **Dynamic format changes on Sender** section below? Otherwise the reader naturally asks _how_ it depends on the Sender itself, and how the Controller can determine it.

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
+
+If keeping the configuration of the Sender, the NMOS Controller MUST `GET /constraints/supported` and `GET /constraints/active` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such processing of them that they can be utilized for building Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, the NMOS Controller SHOULD inform the user about it.
+Sometimes there can be streams compliant with the current active constraints of the Sender, but not compliant with the resulting `constraint_sets`. In these scenarios the NMOS Controller SHOULD inform the user that the Receivers cannot be accommodated without replacing the Active Constraints.
```
garethsb, line 10: I still wonder whether the duplicated steps between the two scenarios isn't helpful.

I've attempted again to extract the commonality to see how that reads.

During the call today, we also removed the sentence that "If the Constraint Sets have been applied successfully, the NMOS Controller SHOULD then make the IS-05 connections of the Receivers", in order to focus the conformance language on Stream Compatibility Management API.

```suggestion
The NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and process them to build Constraint Sets satisfying all (by default) or some (depending on User preferences) of the Receivers, using Parameter Constraints supported by the Sender.
If the Sender does not support all the Parameter Constraints used in the Receiver Capabilities, the NMOS Controller SHOULD inform the User about it.
If the processing of the Receiver Capabilities results in no Constraint Sets, the NMOS Controller SHOULD inform the User that all of the chosen Receivers cannot be satisfied at the same time. 

If changing the configuration of the Sender, the NMOS Controller MUST then `PUT /constraints/active` to the Sender to replace the Active Constraints.

If keeping the configuration of the Sender, the NMOS Controller MUST instead `GET /constraints/active` from the Sender and compare these Active Constraints with the Constraint Sets built from the Receiver Capabilities.
Sometimes there could be streams compliant with the current Active Constraints that would not be compliant with the Constraint Sets built from the Receiver Capabilities.
In these scenarios, the NMOS Controller SHOULD inform the User that compatibility of the Sender's stream with the chosen Receivers cannot be assured without replacing the Active Constraints.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
+After that the NMOS Controller SHOULD make the IS-05 connection of the Receivers.
```
garethsb, line 11: Do we really want a Controller to go ahead and make the connections if it has identified that some of the Receivers cannot be accommodated? Do we want to recommend that the User is involved at this point? (Note that in the "changing the configuration" scenario, the Controller SHOULD make the connection if the constraints have been successfully applied. Why should it be different in the "keeping the configuration" case?)

docs/Behaviour - Client Side.md
garethsb: We found this during the call:

https://github.com/AMWA-TV/is-11/blob/309d430f4475aec9307b7dc02d3a2641eaf793f6/APIs/StreamCompatibilityManagementAPI.raml#L123-L125

docs/Behaviour - Client Side.md
garethsb:
```suggestion

After this the NMOS Controller can make the chosen connections via IS-05.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
+
+After breaking these connections via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender after making it inactive.
```
garethsb, line 13:
```suggestion
Subsequently, after breaking these connections via IS-05, the NMOS Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender after making it inactive.
```

docs/Behaviour - Client Side.md
garethsb:
```suggestion
Before making an [IS-05][IS-05] connection, an NMOS Controller chooses a Sender and a number of Receivers. The NMOS Controller MAY keep the current configuration of the Sender or change its configuration.
```

I'd like to move the explanation of an active Sender responding with `423` (Locked) below, but I think here in the "Client Side" docs, it should be written in a way which describes any requirements on the Controller.

E.g. something like:

"If the Sender returns `423` Locked, the Controller SHOULD deactivate the Sender and try again."

However, there are problems with this. This is another time we might want to inform the User, isn't it? Would a Controller that never reattempted the operation automatically, but required the User to explicitly deactivate the Sender and try again, be valid? On the other hand, would a Controller that always deactivated a Sender _before_ trying to `PUT` Active Constraints be valid? Formulating a (semi)automated test case for this is awkward as a result.

So far as we determined in the call, a Controller cannot determine in advance whether the Sender supports changing configuration while active.

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
+
+In order to change the configuration of the Sender without establishing an IS-05 connection, NMOS Controller MUST use `PUT /constraints/active` endpoint. If the Sender is active, deactivating such Sender MAY be required and depends on the Sender itself.
```
garethsb, line 15: (a) What does the conformance language really achieve here?
(b) Let's not repeat the explanation of deactivation.

```suggestion
The NMOS Controller can also change the configuration of the Sender without establishing an IS-05 connection.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,7 +2,17 @@
...
 
 Constraint Sets of Active Constraints SHOULD represent a consensus among Receivers. For example, there are Receiver A, Receiver B, Receiver C with equal Receiver Capabilities consisting of Constraint Sets 1, 2, 3, 4, 5 and Receiver D with Constraint Sets 2, 3, 4, 5 and 6 which is incompatible with the previous ones. In this case the consensus is Constraint Sets 2, 3, 4, 5 which are common for all the four Receivers. The user may control how the consensus is obtained, providing preferences to the NMOS Controller among the Receivers and the Constraint Sets of each Receiver and criteria about the size of the consensus.
```
garethsb, line 17:
```suggestion
Constraint Sets built from the Receiver Capabilities SHOULD represent a consensus among the Receivers. For example, there are Receiver A, Receiver B, Receiver C with equal Receiver Capabilities consisting of Constraint Sets 1, 2, 3, 4, 5 and Receiver D with Constraint Sets 2, 3, 4, 5 and 6 which is incompatible with the previous ones. In this case the consensus is Constraint Sets 2, 3, 4, 5 which are common for all the four Receivers. The User may control how the consensus is obtained, providing preferences to the NMOS Controller among the Receivers and the Constraint Sets of each Receiver and criteria about the size of the consensus.
```

alabou:
It seems to me that we are still not allowing to change the configuration of the Sender without considering real Receivers ....

garethsb:
How about beginning with "Before making connections, an NMOS Controller can configure the Sender, taking account of the Receivers to which it will be connected."

With this language, I think this is sufficient to cover potential Receivers in the future.

(Might want to change subsequent uses of choose/chosen having changed that here.)

alabou:
"An NMOS Controller MAY configure the Sender before making [IS-05][IS-05] connections." ... Then we need to indicate how this is being performed because we'll be describing it for the case "Before making an [IS-05][IS-05] connection" ...

N-Nagorny:
Closed in favor of #115

PR #115, opened 2023-03-13
Another attempt to refactor the client behaviour section

garethsb:
Based on discussion on #114.

### DIFF
--- a/docs/Behaviour - Client Side.md
+++ b/docs/Behaviour - Client Side.md
@@ -2,16 +2,47 @@
 
 ## Active Constraints of Sender
 
-Before making an [IS-05][IS-05] connection, NMOS Controller chooses a Sender and a number of Receivers. Then NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and make such processing of them that they can be utilized for building Active Constraints satisfying all (by default) or some (depending on user preferences) of the Receivers and using Parameter Constraint URNs supported by the Sender. If the Sender supports fewer Parameter Constraint URNs than used in the Receiver Capabilities, the NMOS Controller SHOULD inform the user about it. If the processing of the Receiver Capabilities results in an empty `constraint_sets` of Active Constraints, NMOS Controller SHOULD inform the user about it. After that NMOS Controller MUST `PUT /constraints/active` to the Sender and SHOULD make the IS-05 connections if the Constraints have been applied succesfully to the Sender. After breaking these connections via IS-05, NMOS Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender.
+An NMOS Controller manages stream compatibility using the Active Constraints of the Sender.
+The Active Constraints of the Sender are expressed as Constraint Sets that restrict the stream the Sender can transmit.
+The Controller MAY build these Constraint Sets from the Receiver Capabilities of one or more Receivers.
 
-Constraint Sets of Active Constraints SHOULD represent a consensus among Receivers. For example, there are Receiver A, Receiver B, Receiver C with equal Receiver Capabilities consisting of Constraint Sets 1, 2, 3, 4, 5 and Receiver D with Constraint Sets 2, 3, 4, 5 and 6 which is incompatible with the previous ones. In this case the consensus is Constraint Sets 2, 3, 4, 5 which are common for all the four Receivers. The user may control how the consensus is obtained, providing preferences to the NMOS Controller among the Receivers and the Constraint Sets of each Receiver and criteria about the size of the consensus.
+### Building the Constraint Sets
 
-`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates relative preference between Constraint Sets of that Receiver. The NMOS Controller MAY use these values to make a decision about how to process these Constraint Sets and determine preference of Constraint Sets of Active Constraints. It SHOULD NOT propagate values of this property as is from Receiver Capabilities to Active Constraints.
+The NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and process them to build Constraint Sets satisfying all (by default) or some (depending on User preferences) of the Receivers, using Parameter Constraints supported by the Sender.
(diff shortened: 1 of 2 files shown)


docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,47 @@
...
+
+If keeping the configuration of the Sender, the Controller MUST instead `GET /constraints/active` from the Sender and evaluate these Active Constraints against the Constraint Sets built from the Receiver Capabilities.
+Sometimes there could be streams compliant with the current Active Constraints that would not be compliant with the Constraint Sets built from the Receiver Capabilities.
+In these scenarios, the Controller SHOULD inform the User that compatibility of the Sender's stream with the chosen Receivers cannot be assured without replacing the Active Constraints.
+
+After this the Controller can make the chosen connections via [IS-05][].
+
+Subsequently, after breaking these connections via IS-05, the Controller is RECOMMENDED to `DELETE /constraints/active` of the Sender after making it inactive.
+
+The Controller can also change the configuration of the Sender without establishing an IS-05 connection.
```
alabou: can => MAY

docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,47 @@
...
+### Building the Constraint Sets
 
-`urn:x-nmos:cap:meta:preference` property in Constraint Sets of Receiver Capabilities indicates relative preference between Constraint Sets of that Receiver. The NMOS Controller MAY use these values to make a decision about how to process these Constraint Sets and determine preference of Constraint Sets of Active Constraints. It SHOULD NOT propagate values of this property as is from Receiver Capabilities to Active Constraints.
+The NMOS Controller MUST `GET /constraints/supported` from the Sender, MUST collect Receiver Capabilities from the Receivers and process them to build Constraint Sets satisfying all (by default) or some (depending on User preferences) of the Receivers, using Parameter Constraints supported by the Sender.
```
alabou: ... still describing only one way of getting active constraints "MUST collect Receiver Capabilities from the Receivers" even if the previous paragraph indicates "The Controller MAY build these Constraint Sets from the Receiver Capabilities of one or more Receivers." which open the possibility of build constraint sets some other way.

It seems to me that we must have something like "constraint sets can be build from Receiver capabilities or from some Controller configurations or user requirements expressed as BCP-004-01 capabilities ..."

SoloAs: Agreeing on this one. Otherwise we are assuming that having Receivers is mandatory, which excludes the case with changing configuration of the Sender without any Receivers intact.

garethsb: You're absolutely right, having a MUST following on from a MAY is bad spec language. I'll rework it.

docs/Behaviour - Client Side.md
garethsb:
```suggestion
The Controller MAY also change the configuration of the Sender without establishing an IS-05 connection.
```

docs/Behaviour - Client Side.md
garethsb: @SoloAs, @alabou, I've had another go...

alabou (approved):

docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,50 @@
...
+The Controller MAY use these values to make a decision about how to process these Constraint Sets and determine preference of Constraint Sets of Active Constraints.
+It SHOULD NOT propagate values of this property as is from Receiver Capabilities to Active Constraints.
+
+The `urn:x-nmos:cap:meta:preference` property in Constraint Sets of Active Constraints indicates relative preference between Constraint Sets of the Active Constraints of the Sender.
+The Controller MAY fill in this property in Constraint Sets based on any additional information about Receivers.
+
+For example, if the User in any way informs the Controller that Receiver D has precedence over the consensus of Receivers A, B, C and `urn:x-nmos:cap:meta:preference` values of Receiver D state that Constraint Set 6 has higher preference than Constraints Sets 2, 3, 4, 5, then the Constraint Sets could be 2, 3, 4, 5, 6 where Constraint Set 6 has the highest preference.
+
+Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely during this process.
```
```suggestion
Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to `false` MUST be ignored completely during this process.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,50 @@
...
+
+### Using the Constraint Sets
+
+If changing the configuration of the Sender, the Controller MUST `PUT /constraints/active` to the Sender to replace the Active Constraints.
+In case of changing the configuration of an active Sender, deactivating the Sender MAY be necessary.
+The Sender indicates the Active Constraints cannot currently be updated with the `423` Locked response.
```
```suggestion
The Sender indicates the Active Constraints cannot currently be updated with the `423` (Locked) response.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,50 @@
...
+
+If keeping the configuration of the Sender, the Controller MUST instead `GET /constraints/active` from the Sender and evaluate these Active Constraints against the Constraint Sets it has built.
```
```suggestion
If keeping the configuration of the Sender, the Controller MUST `GET /constraints/active` from the Sender and evaluate these Active Constraints against the Constraint Sets it has built.
```

docs/Behaviour - Client Side.md
```diff
@@ -2,16 +2,50 @@
...
+Sometimes there could be streams compliant with the current Active Constraints that would not be compliant with the Constraint Sets it has built.
+In these scenarios, the Controller SHOULD inform the User that compatibility of the Sender's stream with the chosen Receivers cannot be assured without replacing the Active Constraints.
```
```suggestion
If the current Active Constraints allow streams, that would not be compliant with the Constraint Sets it has built, the Controller SHOULD inform the User that compatibility of the Sender's stream with the chosen Receivers cannot be assured without replacing the Active Constraints.
```

N-Nagorny (approved):
It's magnificent! Do we want to move `Behaviour - Client Side.md` to `Controllers.md` as IS-04 and -05 does?

SoloAs (approved):

Issue #116, opened 2023-03-13
Dynamic format changes on Sender


garethsb:
I'm struggling a bit with this section.

> # Changes which break Active Constraints
> 
> NMOS Controller MUST track Sender's `subscription`'s `active` property during and after the IS-05 connection. If it became false unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.

* This section is called "Changes which break Active Constraints" but making/breaking connections using IS-05 doesn't change the stream format. In the case of an RTP multicast Sender, making "the IS-05 connection" may not need any requests to the Sender's Connection API at all. What are the "changes" being discussed? The Controller doesn't directly ask for a stream format change, it updates the Active Constraints, and the Sender may change the format in response. Are we talking about enabling an inactive Sender via IS-05 for the first time after changing Active Constraints?
* Why "MUST" the Controller track `subscription.active` if it only SHOULD do anything about anything it discovers as a result?

> # Changes which do not break Active Constraints
> 
> Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.

* The "so" here seems strange. Is it saying that the Controller SHOULD/MUST track the IS-04 `version` instead of/as well as `subscription.active` in order to identify changes in the transport file?

alabou:
The "Changes" that we are referring to here are about the signal flowing through the Source->Flow->Sender .. i.e. Flow attributes changes.

Changes which break Active Constraints => means that the Flow became non-compliant with the active constraints. => active_constraints_violation

Changes which do not break Active Constraints => means that the Flow may have changed but is still compliant with the active constraints. => constrained

"The "so" here seems strange" ... agreed ... the sentence indicate that even if Flow attributes have changed, the Flow may still be compliant with the active constraints, the Sender remains active and the Controller SHOULD patch the Receivers again if the SDP transport file have changed because the Receiver are assumed as not being able to continue streaming if the SDP transport file has changed.

garethsb:
> The "Changes" that we are referring to here are about the signal flowing through the Source->Flow->Sender .. i.e. Flow attributes changes

[...]

PR #117, opened 2023-03-16
Rewrite the section about dynamic updates

N-Nagorny:
Based on #116.

### DIFF
--- a/docs/Behaviour - Client Side.md
+++ b/docs/Behaviour - Client Side.md
@@ -15,9 +15,8 @@ Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set 
 
 ## Dynamic Update of Resources
 
-The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 APIs to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
+The Controller uses Stream Compatibility API in conjunction with IS-04 APIs to discover and track the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
 
-In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
 * The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API and indicate statuses of such Senders/Receivers
 * The Controller MUST report to the User when `state` attribute of the status of a Sender remains `active_constraints_violation` or `no_essence` after a maximum of 30 seconds.
 * The Controller SHOULD report to the User when `state` attribute of the status of an active Receiver becomes `non_compliant_stream`.
--- a/docs/Behaviour - Client Side.md
+++ b/docs/Behaviour - Client Side.md
@@ -13,14 +13,14 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
 Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely while making the processing.
 
 
-## Dynamic format changes on Sender
+## Dynamic Update of Resources
 
-### Changes which break Active Constraints
+The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 Query API (if available) or IS-04 Node API to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
 
-NMOS Controller MUST track Sender's `subscription`'s `active` property during and after the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
-
-### Changes which do not break Active Constraints
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+* The Controller MUST indicate statutes of Inputs, Outputs, Senders and Receivers.
+* The Controller MUST report to the User when `status` attribute of the state of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence` or `awaiting_essence`.
+* The Controller SHOULD report to the User when `status` attribute of the state of an active Receiver becomes `non_compliant_stream`.
+* The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
+* The Controller MUST reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
 
 [IS-05]: https://specs.amwa.tv/is-05/
--- a/docs/Behaviour - Client Side.md
+++ b/docs/Behaviour - Client Side.md
@@ -18,10 +18,11 @@ Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set 
 The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 APIs to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
 
 In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
-* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers.
-* The Controller MUST report to the User when `status` attribute of the state of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence` or `awaiting_essence`.
-* The Controller SHOULD report to the User when `status` attribute of the state of an active Receiver becomes `non_compliant_stream`.
-* The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
-* The Controller MUST reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
+* The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API and indicate statuses of such Senders/Receivers
+* The Controller MUST report to the User when `state` attribute of the status of a Sender remains `active_constraints_violation` or `no_essence` after a maximum of 30 seconds.
+* The Controller SHOULD report to the User when `state` attribute of the status of an active Receiver becomes `non_compliant_stream`.
+* The Controller SHOULD be capable of performing `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
+* The Controller SHOULD indicate statuses of Inputs and Outputs and indicate the Inputs associated to a given Sender and Outputs associated to a given Receiver for Senders/Receivers listed in Stream Compatibility APIs.
+* The Controller SHOULD reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
 
 [IS-05]: https://specs.amwa.tv/is-05/
--- a/docs/Behaviour - Client Side.md
+++ b/docs/Behaviour - Client Side.md
@@ -15,9 +15,10 @@ Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set 
 
 ## Dynamic Update of Resources
 
-The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 Query API (if available) or IS-04 Node API to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
+The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 APIs to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
 
-* The Controller MUST indicate statutes of Inputs, Outputs, Senders and Receivers.
+In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
+* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers.
 * The Controller MUST report to the User when `status` attribute of the state of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence` or `awaiting_essence`.
 * The Controller SHOULD report to the User when `status` attribute of the state of an active Receiver becomes `non_compliant_stream`.
 * The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -275,7 +275,7 @@ mediaType: application/json
                 application/octet-stream:
                   type: file
             204:
-              description: There is no Base EDID set or the Input doesn't support Base EDID at all.
+              description: There is no Base EDID set or the Input doesn't support Base EDID or EDID at all.
             307:
               description: A temporary redirect to the Base EDID file.
             404:
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -275,7 +275,7 @@ mediaType: application/json
                 application/octet-stream:
                   type: file
             204:
-              description: There is no Base EDID set.
+              description: There is no Base EDID set or the Input doesn't support Base EDID at all.
             307:
               description: A temporary redirect to the Base EDID file.
             404:
@@ -301,7 +301,7 @@ mediaType: application/json
               body:
                 type: !include schemas/error.json
             405:
-              description: Returned when the Input doesn't support changing Base EDID or doesn't support EDID at all.
+              description: Returned when the Input doesn't support Base EDID or doesn't support EDID at all.
               body:
                 type: !include schemas/error.json
             423:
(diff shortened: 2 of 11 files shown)


docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,14 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
 Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely while making the processing.
 
 
-## Dynamic format changes on Sender
+## Dynamic Update of Resources
 
-### Changes which break Active Constraints
+The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 Query API (if available) or IS-04 Node API to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
 
-NMOS Controller MUST track Sender's `subscription`'s `active` property during and after the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
```
alabou, line 20: You removed the "tracking sender subscription active property" aspect ... We must favor monitoring through the registry query/subscription mechanism first ... I don't think it would be a good idea for Controller to poll the status/state. In all aspects of Controller monitoring we must make sure that it is being performed not by polling the device but the registry.

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,14 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
...
-
-### Changes which do not break Active Constraints
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+* The Controller MUST indicate statutes of Inputs, Outputs, Senders and Receivers.
```
alabou: "statutes" ???

There are a number of new MUST clauses. Are there any other API requiring the Controller to "indicate" the state of something?

garethsb: Let's composer with the language in e.g. https://specs.amwa.tv/is-05/releases/v1.1.2/docs/Controllers.html#connection-management

("statuses" I guess!)

N-Nagorny:
```suggestion
* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers.
```

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,14 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
 Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely while making the processing.
 
 
-## Dynamic format changes on Sender
+## Dynamic Update of Resources
 
-### Changes which break Active Constraints
+The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 Query API (if available) or IS-04 Node API to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
```
```suggestion
The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 APIs to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.

In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
```

docs/Behaviour - Client Side.md
N-Nagorny: If I correctly understood Gareth in #116, it's implied that Controller tracks resources' properties, we need to dictate what Controller should do with their values. Moreover, `active` property update is a consequence of changing the State of a Sender to `active_constraints_violation` so maybe this is what Controller should track instead.

[The current text](https://specs.amwa.tv/is-11/branches/v1.0-dev/docs/Interoperability.html#version-increments) says:
>When State of Sender or Receiver is changed, or when the set of Inputs/Outputs associated with the Sender/Receiver is changed, then the version attribute of the relevant IS-04 Sender or Receiver MUST be incremented.

Do I correctly understand, that it means that Controller doesn't have to periodically poll `/status` and can it do that only after gets `version` update from Query API?

>In all aspects of Controller monitoring we must make sure that it is being performed not by polling the device but the registry.

I tried to express it with `The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 Query API (if available) or IS-04 Node API`. [IS-04 RECOMMENDs](https://specs.amwa.tv/is-04/releases/v1.3.2/docs/Controllers.html#websocket-subscriptions) to use WebSocket subscriptions. Inserted this paragraph as a suggestion.

alabou: "Do I correctly understand, that it means that Controller doesn't have to periodically poll /status and can it do that only after gets version update from Query API?"

What the text says is a requirement for the Node.

"In all aspects of Controller monitoring we must make sure that it is being performed not by polling the device but the registry."

Where is this sentence coming from?

At the IS-11 level we must make sure people will not mis-interpret our specification because they haven't read the full IS-04 and/or Controller specifications and emphasis when possible that because an IS-11 Node have a requirement to update the Sender's version is various cases, the Controller SHOULD monitor from the Registry when available.

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,15 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
...
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
+* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers.
+* The Controller MUST report to the User when `status` attribute of the state of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence` or `awaiting_essence`.
+* The Controller SHOULD report to the User when `status` attribute of the state of an active Receiver becomes `non_compliant_stream`.
+* The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
+* The Controller MUST reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
```
alabou: Are you referring to Resources appearing in the /inputs, /outputs, /senders/*inputs or /receivers/*/outputs (id come and go) or to the state (connected, no_signal, ...)?

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,15 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
...
 
-NMOS Controller MUST track Sender's `subscription`'s `active` property during and after the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
-
-### Changes which do not break Active Constraints
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
+* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers.
```
alabou: If there is no inputs, outputs the Controller does not have any requirement, not even to indicate that there is none. Same thing when no input are associated to a Sender and no output is associated to a Receiver ... For Nodes not supporting the concept of input/output we must make sure the Controller has no requirement about something not supported by the Node.

docs/Behaviour - Client Side.md
N-Nagorny: The last item here is about showing Inputs/Outputs presence (`/inputs/{inputId}/properties` and `/outputs/{outputId}/properties`). A Controller, that shows only IDs coming and going, satisfies this requirement. Showing states is covered by the item before the first item here (not included to your quote but it's the 1st in this list).

There is no requirement about showing relationship between Inputs/Outputs with Senders/Receivers though. Maybe we have to add it as well.

```suggestion
* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers and relationship (or its absence) between Inputs/Outputs and Senders/Receivers listed in Stream Compatibility APIs.
* The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API
* The Controller MUST report to the User when `status` attribute of the state of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence` or `awaiting_essence`.
* The Controller SHOULD report to the User when `status` attribute of the state of an active Receiver becomes `non_compliant_stream`.
* The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
* The Controller MUST reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
```

docs/Behaviour - Client Side.md
N-Nagorny: Tried to cover it in the suggestion below.

docs/Behaviour - Client Side.md
alabou: awaiting_essence is a transient state and MUST not be reported to the user.

alabou: "when `status` attribute of the state of a Sender" => seems to be "when `state` attribute of the status of a Sender"

N-Nagorny:
```suggestion
* The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers and relationship (or its absence) between Inputs/Outputs and Senders/Receivers listed in Stream Compatibility APIs.
* The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API
* The Controller MUST report to the User when `state` attribute of the status of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence`.
* The Controller SHOULD report to the User when `state` attribute of the status of an active Receiver becomes `non_compliant_stream`.
* The Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
* The Controller MUST reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
```
Removed `awaiting_essence`. Should we dictate that Controller MUST not report about it?

alabou: "The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers and relationship (or its absence) between Inputs/Outputs and Senders/Receivers listed in Stream Compatibility APIs."

It seems to me that "relationship" is not precise enough. I would phrase it as follow instead:

"The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers and SHOULD indicate the Inputs associated to a given Sender and Outputs associated to a given Receiver for Senders/Receivers listed in Stream Compatibility APIs."

The word "relationship" in general could have meant to show the interconnection of inputs with senders in general and outputs with receivers in general and I think that this is out of scope for IS-11.

alabou: "* The Controller MUST report to the User when `state` attribute of the status of a Sender becomes `active_constraints_violation` and SHOULD report when it becomes `no_essence`."

I would be careful here and not use MUST here but a SHOULD because of the "volatility" of the `active_constraints_violation` state .. For example when changing the EDID preferred mode the source does not react instantaneously such that the transition may not be clean and the Sender may get to the `active_constraints_violation` state. The Controller in such cases SHOULD restart the Sender automatically without telling the User because it may be a temporary state and a subsequent restart would finally remain stable with the state `constrained`.

To me the following text would be more appropriate.

"* The Controller SHOULD report to the User when `state` attribute of the status of a Sender becomes `active_constraints_violation`  or `no_essence`."

Otherwise this would also be acceptable:

"* The Controller MUST report to the User when `state` attribute of the status of a Sender remains `active_constraints_violation` or  `no_essence` after a maximum of 30 seconds."

N-Nagorny:
```suggestion
* The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API and indicate statuses of such Senders/Receivers
* The Controller MUST report to the User when `state` attribute of the status of a Sender remains `active_constraints_violation` or `no_essence` after a maximum of 30 seconds.
* The Controller SHOULD report to the User when `state` attribute of the status of an active Receiver becomes `non_compliant_stream`.
* The Controller SHOULD be capable of performing `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
* The Controller SHOULD indicate statuses of Inputs and Outputs and indicate the Inputs associated to a given Sender and Outputs associated to a given Receiver for Senders/Receivers listed in Stream Compatibility APIs.
* The Controller SHOULD reflect changes in presence/absence of Inputs and Outputs to the User after a maximum of 30 seconds.
```
Added the 30 seconds timeout for reporting about `state` of a Sender.

N-Nagorny: >The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers and SHOULD indicate the Inputs associated to a given Sender and Outputs associated to a given Receiver for Senders/Receivers listed in Stream Compatibility APIs.

It may be read as if orphan Inputs/Outputs can be ignored at all. Does the following sound more precise?
>The Controller MUST indicate statuses of Inputs, Outputs, Senders and Receivers listed in Stream Compatibility APIs and SHOULD indicate Inputs associated to a given Sender and Outputs associated to a given Receiver for such Senders/Receivers.

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,16 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
...
-
-### Changes which do not break Active Constraints
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
+* The Controller MUST indicate whether each Sender/Receiver discovered via IS-04 APIs is present in Stream Compatibility API and indicate statuses of such Senders/Receivers
+* The Controller MUST report to the User when `state` attribute of the status of a Sender remains `active_constraints_violation` or `no_essence` after a maximum of 30 seconds.
+* The Controller SHOULD report to the User when `state` attribute of the status of an active Receiver becomes `non_compliant_stream`.
+* The Controller SHOULD be capable of performing `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation when the related active Sender updates the transport file.
```
N-Nagorny, line 23: This requires the Controller to act on events other than user input. `SHOULD` may be too strong for it.

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,16 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
 Constraint Sets of Receiver Capabilities with `urn:x-nmos:cap:meta:enabled` set to false MUST be ignored completely while making the processing.
 
 
-## Dynamic format changes on Sender
+## Dynamic Update of Resources
 
-### Changes which break Active Constraints
+The Controller MUST be capable of using Stream Compatibility API in conjunction with IS-04 APIs to discover and dynamically report the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
```
```suggestion
The Controller uses Stream Compatibility API in conjunction with IS-04 APIs to discover and track the state of any relevant resources, including Inputs, Outputs, Senders and Receivers.
```

alabou: Should it be more specific "The Controller uses Stream Compatibility API in conjunction with IS-04 Query API to discover and track the state of any relevant resources, including Inputs, Outputs, Senders and Receivers." because IS-04 says "The Controller uses the Query API to discover and track the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver."

docs/Behaviour - Client Side.md
alabou: I'm not sure I understand which SHOULD is too strong ... could you be more precise?

docs/Behaviour - Client Side.md
```diff
@@ -13,14 +13,16 @@ Constraint Sets of Active Constraints SHOULD represent a consensus among Receive
...
 
-NMOS Controller MUST track Sender's `subscription`'s `active` property during and after the IS-05 connection. If it became `false` unexpectedly, NMOS Controller SHOULD `GET /status` to figure out the reason of breaking the connection. If the State of the Sender is `active_constraints_violation`, then NMOS Controller SHOULD either disconnect all the Receivers or `PUT /constraints/active` once again and re-enable Sender via IS-05.
-
-### Changes which do not break Active Constraints
-
-Sender's changes which meet the Active Constraints do not affect the `active` property so if the transport file of the Sender has changed, the NMOS Controller SHOULD `PATCH /staged` of all the connected Receivers with the new `transport_file` requesting immediate activation.
+In order to avoid polling of the Stream Compatibility API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism of IS-04 Query API.
```
N-Nagorny: Should we align this text with https://specs.amwa.tv/is-05/releases/v1.1.2/docs/Controllers.html#version-timestamp?

docs/Behaviour - Client Side.md
N-Nagorny: The last one.

docs/Behaviour - Client Side.md
N-Nagorny: There is no Query API in the Peer to Peer mode and I tried to include this case. I will raise it in the IS-04 PR.

docs/Behaviour - Client Side.md
```suggestion
```

PR #123, opened 2023-04-24
Revisit edid.json

N-Nagorny:
I looked at the JSON Schema for EDID and found that it doesn't conform the EDID spec. This PR fixes it.

Not sure what to do with Detailed Timing Descriptions, Standard Timings and `cta_861_timings`. The assumption was that they all may be described with `edid_timing.json` consisting of `width`, `height`, and `frame_rate` but it's more complex than that. Do we need to describe them in JSON at all?


garethsb:
Just wondering... do Controllers need the EDID-as-JSON at all? What purpose is it serving?

N-Nagorny:
There is a User Story behind the activity.

> As a Pro AV digital signage installer, I need to colormatch the monitors on my video wall. To do this, I need to read the detailed EDID information, which includes the monitor's color profile, model number and other such information, which I use in my software for callibration.

From the other side, there is another User Story, and it's not quite clear whether they should stay two separate ones.

> As a Pro AV installer, I have developed some in house EDID software that reads the more detailed data from the monitor, such serial numbers, manufacture date and color management information. When I make video walls for customers, I use IPMX AV over IP hardware to deliver content. I would like to be able to retrieve the EDID binary representation from my monitors into my home-built software, so that I can use that information to adjust the monitors and manage my system.

N-Nagorny:
If we remove the JSON representation of EDID from the spec and after some time understand that we need some EDID properties simply provided to a Controller, maybe BCP-004-01 is capable of covering the most of the EDID properties. BCP-005-01 already defines such recommendations, and it's already a spec where a Node is implied to parse EDID to JSON for Controllers.

garethsb:
That's an interesting point. If BCP-005-01 provides sufficient coverage for mapping _Output_ EDID into Receiver caps Parameter Constraints that are already defined, then yes, one might assume that sufficient information is also provided by Source, Flow and Sender attributes to cover _Input_ EDID. And if we identify something missing, maybe it could be handled by adding new Parameter Constraints and Source/Flow/Sender attributes which can be added to the Parameter Registers?

N-Nagorny:
Yes, this is what I mean. Regarding covering _Input_ EDID, it looks less important than covering _Output_ EDID.

The first User Story gives colormatching as an example. If the installer faces troubles, he can check the Receiver Caps and see `urn:x-nmos:cap:meta:hdmi_sink_model_number` or so and `urn:x-nmos:cap:format:edid_gamma` or so and tweak a baseband source based on it.

There are no User Stories when the end user needs to know what's in the _Input_ EDID. If it's possible to say, it's an implementation detail even more than the _Output_ EDID.

garethsb:
> he can check the Receiver Caps and see `urn:x-nmos:cap:meta:hdmi_sink_model_number` or so and `urn:x-nmos:cap:format:edid_gamma`

Just checking... these are examples of things that don't exist yet, but could be added later? Or do we need to support these to provide for this User story before elevation of BCP-005-01?

N-Nagorny:
I've given them as examples for further development and then had a discussion with Andrew Starks. If I understood him correctly, BCP-005-01 mostly covers what we need for the User Stories, so I'll make commits removing JSON representation of EDID.

garethsb:
I think there are some more references e.g. in Overview?

N-Nagorny:
Hm, I see only references to EDID binaries like this.

> Each Input has a `properties` resource, which has metadata about this Input, and an `edid` resource, which may hold EDIDs associated with this Input.

garethsb:
You're right, I think I misinterpreted it.

garethsb (approved):
LGTM

PR #129, opened 2023-05-16
Make IS-11 schemas self-contained

### LINKED ISSUE #125: Should we make IS-11 schemas self contained?
https://github.com/AMWA-TV/is-11/blob/ff335ee226b8190ae5d79b4d81936e338ec2f95e/APIs/schemas/constraints_active.json#L13

https://github.com/AMWA-TV/is-11/blob/ff335ee226b8190ae5d79b4d81936e338ec2f95e/APIs/schemas/input.json#L7

https://github.com/AMWA-TV/is-11/blob/ff335ee226b8190ae5d79b4d81936e338ec2f95e/APIs/schemas/output.json#L7

garethsb:
Would resolve #125


garethsb:
Should also remove or change the `Resource-Core-Schema` reference in _Overview.md_ that currently points directly at rendered IS-04 v1.3.2 _resource_core.json_ schema to just say that Inputs and Outputs share the same core resource attributes as IS-04 resources. After merging #131.

N-Nagorny:
It would be good to emphasize which IS-04 version IS-11 shares this same JSON Schema with, and something similar for BCP-004-01.

garethsb:
The interoperability section does already mention IS-04 v1.2 and state compatibility with BCP-004-01 v1.0. Maybe that's sufficient?

N-Nagorny:
As an implementer, I open the interoperability section to figure out dependencies, and they may be not very precise (`version 1.2 or greater`). As a spec maintainer, I may want to know where parts of the spec were exactly copied from. Maybe creating `README.md` in `APIs/schemas` could help?

N-Nagorny:
@garethsb, is it an option to add a BCP-004-01 submodule to the repo? It simplifies providing repo content to air-gapped tools, excludes possibilities for contradictions between copied and original files and resolves the issue of tracking where files were taken from.

alabou:
I would prefer a single instance of the schemas to prevent duplicating information. This is up to the NMOS organisation to decide of the policy for handling shared schemas ... So if the NMOS rule is to duplicate, I'm ok with that.

alabou (approved):

PR #130, opened 2023-05-16
Draft IS-13-like mechanism to update distinguishing info of Inputs and Outputs

### LINKED ISSUE #126: Should distinguishing info of Inputs and Outputs be PATCHable?
See https://specs.amwa.tv/is-13 for background and https://github.com/AMWA-TV/is-13/issues/17.

garethsb:
For discussion!
Would resolve #126

### DIFF
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -252,6 +252,36 @@ mediaType: application/json
             description: Returned when the requested Input ID does not exist.
             body:
               type: !include schemas/error.json
+      options:
+        description: A pre-flight check generally used for Cross-Origin Resource Sharing (CORS) purposes
+        responses:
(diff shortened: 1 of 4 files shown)


garethsb:
An alternative now that IS-13 Annotation API has introduced a `/node/` base path for the Node API-related resources, would be to add a `/streamcompatibility/` base path to that API, with `inputs/` and `/outputs` under it.

alabou:
Why an IS-13 feature is implemented in the IS-11 API?

garethsb:
Can the Stream Compatibility Management activity group first evaluate the feature, whether there are user stories for assigning user annotations to IS-11 inputs and outputs? Then e can decide where it belongs.

peterbrightwell:
If we're going to start reusing examples perhaps have better text than "that one" and "over there" :)

PR #134, opened 2023-06-30
Add "adjust_to_caps" query parameter

### LINKED ISSUE #132: Modifiying Base EDID on its way to become Effective EDID
> Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
>
> If Base EDID is not set, Effective EDID is built on the basis of a default Base EDID defined for the Input by the manufacturer.

Although these paragraphs say that Effective EDID is Base EDID (or vendor's EDID) modified by Active Constraints, there is a nuance here. If Base EDID is not set by a user, the vendor's EDID, that is used instead, stores information about capabilities of the Input. When Base EDID is set, these capabilities are lost. Sometimes, we want to lose them and set exactly what the user wants. In this case there is a risk that a source will start giving unsupported signal, so we may want to modify Base EDID to fix that. We might solve it via PUTting Active Constraints, but we don't know how to build them because the capabilities of the Input are lost. In this case, it may be beneficial to let Input itself modify Base EDID based on its capabilities before applying Active Constraints to Base EDID.

For me, it looks like a flag for PUT `/edid/base` operation that specifies whether an Input is allowed to modify Base EDID if Active Constraints are not set. Its implementation may consist of two parts:
1. An Input's property like `base_edid_support` (`base_edid_alterable`?) showing whether this Input supports such a flag.
2. A flag itself implemented as an additional request header or as a `multipart/form-data`'s property or a property of another resource like `/inputs/{inputId}/properties`.

N-Nagorny:
Resolves https://github.com/AMWA-TV/is-11/issues/132

### DIFF
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -296,11 +296,13 @@ mediaType: application/json
           queryParameters:
             adjust_to_caps:
               description: >
-                If true, the Input is allowed to adjust Base EDID to internal capabilities.
-                If omitted for an Input that supports it, the previous value is preserved.
+                Sets the value of "adjust_to_caps" Input's property.
+                If omitted, the value of the property is preserved.
+                If an implementation doesn't support the property,
+                this query parameter is always ignored.
               type: boolean
               required: false
-              example: adjust_to_caps=true
+              example: true
           body:
             application/octet-stream:
               type: file
--- a/APIs/StreamCompatibilityManagementAPI.raml
+++ b/APIs/StreamCompatibilityManagementAPI.raml
@@ -293,6 +293,14 @@ mediaType: application/json
             403:
         put:
           description: Set Base EDID
+          queryParameters:
+            adjust_to_caps:
+              description: >
+                If true, the Input is allowed to adjust Base EDID to internal capabilities.
+                If omitted for an Input that supports it, the previous value is preserved.
+              type: boolean
+              required: false
+              example: adjust_to_caps=true
           body:
             application/octet-stream:
               type: file
--- a/docs/Behaviour - Server Side.md
+++ b/docs/Behaviour - Server Side.md
@@ -86,13 +86,19 @@ The `/properties` endpoint shows properties of the Input.
 
 An Input MAY support Base EDID and MUST indicate it via `base_edid_support` property. The API documentation reflects how responses from `/edid/base` depend on this property.
 
+Inputs that support Base EDID MAY additionally support adjusting Base EDID to internal capabilities (e.g. capabilities of an Input itself and Senders associated with this Input) via `adjust_to_caps` query parameter of a `PUT /edid/base` request. An Input MUST indicate support of `adjust_to_caps` query parameter via `adjust_to_caps` property presence in `GET /properties` responses.
+
 There is no Base EDID at the initial state. If the Base EDID for an Input changes, then all Senders associated with this Input MUST update their versions (in registered mode this MUST update the registered resources).
 
 ### Effective EDID
 
-Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `false` or not supported, the Effective EDID is equal to the Base EDID.
 
-If Base EDID is not set, Effective EDID is built on the basis of a default EDID defined for the Input by the manufacturer.
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `true`, the Effective EDID is a combination of the Base EDID and the internal capabilities.
+
+If Active Constraints of at least one Sender associated with this Input are not empty, the Effective EDID is such combination of the Base EDID, internal capabilities and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+
+If the Base EDID is not set, the Effective EDID is built on the basis of a default EDID defined for the Input by the manufacturer.
 
 The `/edid/effective` endpoint allows a client to download the Effective EDID if it exists. If the Effective EDID for the Input changes and it is associated with any Senders, then all of the Senders in question MUST update their versions (in registered mode this MUST update the registered resources).
 
--- a/docs/Behaviour - Server Side.md
+++ b/docs/Behaviour - Server Side.md
@@ -94,9 +94,9 @@ There is no Base EDID at the initial state. If the Base EDID for an Input change
 
 If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `false` or not supported, the Effective EDID is equal to the Base EDID.
 
-If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `true`, the Effective EDID is a combination of the Base EDID and the internal capabilities.
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is supported and equal to `true`, the Effective EDID is a combination of the Base EDID and the internal capabilities.
 
-If Active Constraints of at least one Sender associated with this Input are not empty, the Effective EDID is such combination of the Base EDID, internal capabilities and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+If Active Constraints of at least one Sender associated with this Input are not empty, the Effective EDID is a combination of the Base EDID, internal capabilities and Active Constraints of all Senders associated with this Input such that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
 
 If the Base EDID is not set, the Effective EDID is built on the basis of a default EDID defined for the Input by the manufacturer.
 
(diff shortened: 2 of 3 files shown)


docs/Behaviour - Server Side.md
```diff
@@ -83,6 +83,8 @@ The `/properties` endpoint shows properties of the Input. In particular, it MAY
 
 ### Base EDID
 
+An Input MAY support adjusting Base EDID to its internal capabilities via `adjust_to_caps` query parameter of a `PUT /edid/base` request.
```
alabou: Please clarify this .... is the base EDID modified on PUT such that a Get of the base EDID would return the modified base EDID ... or without active constraint the base EDID is modified to become the effective EDID ... etc ...

docs/Behaviour - Server Side.md
```diff
@@ -86,13 +86,19 @@ The `/properties` endpoint shows properties of the Input.
...
+
 There is no Base EDID at the initial state. If the Base EDID for an Input changes, then all Senders associated with this Input MUST update their versions (in registered mode this MUST update the registered resources).
 
 ### Effective EDID
 
-Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `false` or not supported, the Effective EDID is equal to the Base EDID.
+
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `true`, the Effective EDID is a combination of the Base EDID and the internal capabilities.
```
alabou: "is equal to `true`" ===>> "is equal to `true` and supported"

Even if the first case already covers it ...

docs/Behaviour - Server Side.md
```diff
@@ -86,13 +86,19 @@ The `/properties` endpoint shows properties of the Input.
...
+
+If Active Constraints of at least one Sender associated with this Input are not empty, the Effective EDID is such combination of the Base EDID, internal capabilities and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
```
alabou: "this Input that the baseband signal" ===>>> "this Input such that the baseband signal"

docs/Behaviour - Server Side.md
```suggestion
If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is supported and equal to `true`, the Effective EDID is a combination of the Base EDID and the internal capabilities.
```

docs/Behaviour - Server Side.md
```suggestion
If Active Constraints of at least one Sender associated with this Input are not empty, the Effective EDID is a combination of the Base EDID, internal capabilities and Active Constraints of all Senders associated with this Input such that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
```

docs/Behaviour - Server Side.md
```diff
@@ -86,13 +86,19 @@ The `/properties` endpoint shows properties of the Input.
...
 
+Inputs that support Base EDID MAY additionally support adjusting Base EDID to internal capabilities (e.g. capabilities of an Input itself and Senders associated with this Input) via `adjust_to_caps` query parameter of a `PUT /edid/base` request. An Input MUST indicate support of `adjust_to_caps` query parameter via `adjust_to_caps` property presence in `GET /properties` responses.
+
 There is no Base EDID at the initial state. If the Base EDID for an Input changes, then all Senders associated with this Input MUST update their versions (in registered mode this MUST update the registered resources).
 
 ### Effective EDID
 
-Effective EDID is such combination of Base EDID and Active Constraints of all Senders associated with this Input that the baseband signal from the Input can be transmitted by the Senders without breaking Active Constraints.
+If Active Constraints of all Senders associated with this Input are empty and `adjust_to_caps` is equal to `false` or not supported, the Effective EDID is equal to the Base EDID.
```
garethsb, line 95: I assume "the Effective EDID is equal to the Base EDID" is a shorthand for the actual behaviour, given further down "If the Base EDID is not set, the Effective EDID is built on the basis of a default EDID defined for the Input by the manufacturer."

APIs/StreamCompatibilityManagementAPI.raml
```diff
@@ -293,6 +293,14 @@ mediaType: application/json
             403:
         put:
           description: Set Base EDID
+          queryParameters:
+            adjust_to_caps:
+              description: >
+                If true, the Input is allowed to adjust Base EDID to internal capabilities.
+                If omitted for an Input that supports it, the previous value is preserved.
```
garethsb: I don't 100% understand either sentence.

If true, the Input is allowed to adjust... but doesn't have to?

If omitted (or false?), the previous value is preserved... Which previous value? The previous `adjust_to_caps` value, or the previous Base EDID? Wouldn't that be an error?

APIs/StreamCompatibilityManagementAPI.raml
```diff
@@ -293,6 +293,14 @@ mediaType: application/json
...
+              type: boolean
+              required: false
+              example: adjust_to_caps=true
```
garethsb: Example should probably just be `true`?

garethsb (commented):
(I'm sorry, I'm writing this on phone so please forgive if this is a bit messed up)

I'm really very confused about the relationship between the `adjust_to_caps` property and the `adjust_to_caps` query parameter. I'm not at all clear whether the property is a capability of the API, and the query parameter can only be specified when the property is true. The property could be `adjust_to_caps_supported` to make that clear if so?

I'm not clear whether the API is permitted to ignore the query parameter... sounds like it is?
Or should there be some words about rejecting a `PUT` request with the query parameter with `4xx` or `5xx` in some cases?

I think `"adjust_to_caps": true` mean that adjustment is supported? I.e. not that it will always be performed? If so, is `"adjust_to_caps": false` therefore the same as omitting the property?

Even if the property is `true` and the query parameter is `true` the API is permitted to not adjust the Base EDID?

Does adjusting the Base EDID at `PUT` mean that the `GET` returns the adjusted one? Or only the Effective EDID?

garethsb:
What is the expected behaviour **before** this PR if a client `PUT`s a Base EDID which is incompatible with the API implementation in some way? Reject the `PUT`?

APIs/StreamCompatibilityManagementAPI.raml
N-Nagorny: It looks logical to me too, but I used https://github.com/AMWA-TV/is-10/blob/v1.0.x/APIs/AuthorizationAPI.raml#L65 as a reference. Though I checked it again and that file doesn't seem consistent in terms of such examples. https://raml.org/developers/raml-100-tutorial#enter-query-parameters says it should be rather `true`.

APIs/StreamCompatibilityManagementAPI.raml
N-Nagorny: You're right, it should be "the Input adjusts Base EDID to internal capabilities".

Maybe it makes sense to move the functional description to the JSON Schema's property and say here that this query parameter is a setter for that property.

docs/Behaviour - Server Side.md
N-Nagorny: Hm, these two clauses seems not duplicating each other if you meant it. The first three paragraphs of the section describe how the Effective EDID should be built using the Base EDID and the fourth one describes what "If the Base EDID is not set".

N-Nagorny:
@garethsb, thank you for good questions.

> I'm really very confused about the relationship between the adjust_to_caps property and the adjust_to_caps query parameter.

`adjust_to_caps` property is present if an implementation supports this functionality. In this case the query parameter is used to set the value of this property. Otherwise, it's absent and the implementation behaves like before this PR.

> I'm not clear whether the API is permitted to ignore the query parameter... sounds like it is?
Or should there be some words about rejecting a PUT request with the query parameter with 4xx or 5xx in some cases?

Yes, it is. I don't see options to reject a PUT with the query parameter. If an implementation doesn't support this, then it just ignores query parameters. If there are no requirements in other NMOS specs to respond with an error on unexpected query params in a request, it makes sense to preserve such behaviour.

> I think "adjust_to_caps": true mean that adjustment is supported?

It means that the Effective EDID is the last Base EDID set + internal caps. If there is no the Base EDID, then it means that the next Base EDID set without query params will be combined with the internal caps to form the Effective EDID.

> Even if the property is true and the query parameter is true the API is permitted to not adjust the Base EDID?

Maybe it makes this useless so I'll change the language.

> Does adjusting the Base EDID at PUT mean that the GET returns the adjusted one? Or only the Effective EDID?

Only the Effective EDID to let the user compare what they PUT and what they have after applying the internal caps.

> What is the expected behaviour before this PR if a client PUTs a Base EDID which is incompatible with the API implementation in some way? Reject the PUT?

It returns 400 "when the Base EDID is rejected due to validation failure". There's no room for other incompatibility because the Output's EDID may be really bad even in terms of the EDID specification but it may be what the baseband source expects for, so we have to provide the source with it.

APIs/StreamCompatibilityManagementAPI.raml
N-Nagorny: Fixed. 49a62d311d515695754ea0022b17d3cb7acda7c4

APIs/StreamCompatibilityManagementAPI.raml
N-Nagorny: Removed the functional description from the RAML spec. 49a62d311d515695754ea0022b17d3cb7acda7c4

PR #135, opened 2023-07-27
Client behaviour improvements

N-Nagorny:


docs/Behaviour - Client Side.md
```diff
@@ -27,17 +27,17 @@ It SHOULD NOT propagate values of this property as is from Receiver Capabilities
...
 
 ### Using the Constraint Sets
 
 If changing the configuration of the Sender, the Controller MUST `PUT /constraints/active` to the Sender to replace the Active Constraints.
 In case of changing the configuration of an active Sender, deactivating the Sender MAY be necessary.
-The Sender indicates the Active Constraints cannot currently be updated with the `423` (Locked) response.
+The Sender indicates that the Active Constraints cannot currently be updated with the `423` (Locked) response.
 
-If keeping the configuration of the Sender, the Controller MUST `GET /constraints/active` from the Sender and evaluate these Active Constraints against the Constraint Sets it has built.
+If keeping the configuration of the Sender, the Controller MUST `GET /constraints/active` from the Sender and provide the User with the Active Constraints and the Constraint Sets it has built or with the results of evaluation the former against the latter if the Controller is able to do such evaluation.
```
alabou: This language will prevent a Controller to perform autonomous work as it MUST always provide to the User ... This is bad.

N-Nagorny: Agree.

alabou: I still prefer to keep the original text and changing the MUST for a SHOULD if the objective is to minimize the requirements of the Controller.

"If keeping the configuration of the Sender, the Controller SHOULD GET /constraints/active from the Sender and evaluate these Active Constraints against the Constraint Sets it has built. If the current Active Constraints allow streams, that would not be compliant with the Constraint Sets it has built, the Controller SHOULD inform the User that compatibility of the Sender’s stream with the chosen Receivers cannot be assured without replacing the Active Constraints."

